PR #23, opened 2016-10-21
Add Citations to C3C5 and C6 files

ssiccha:
Eamonn is currently looking through the code with us to figure out where the 'undocumented' algorithms come from.


fingolfin:
Thanks for this -- And sorry for the delay, but I overlooked it before. Best to send me a message via slack or email, I might overlook package PRs otherwise; sadly GitHub notifications suck.

PR #27, opened 2017-12-12
Annotate some unused functions/files

ssiccha:
When Eamon O'Brien visited us we had the impression, that these functions/files were not used. @fingolfin do you think that is still true?

### DIFF
--- a/gap/almostsimple.gi
+++ b/gap/almostsimple.gi
@@ -769,6 +769,7 @@ RECOG.DegreeAlternating := function (orders)
     return [degs, prims];
 end;
 
+# FIXME: dead code? (it's callers are all dead)
 RECOG.RecognizeAlternating := function (orders)
     local   tmp,  degs,  prims,  mindeg,  p1,  p2,  i;
    tmp := RECOG.DegreeAlternating (orders);
--- a/gap/matimpr.gi
+++ b/gap/matimpr.gi
@@ -13,6 +13,7 @@
 ##  This should better be (and is now) called "LowIndex".
 ##
 #############################################################################
+## TODO: WHOLE FILE NOT USED? Was probably replaced by d247.gi
 
 #############################################################################
 ##
--- a/gap/tensor.gi
+++ b/gap/tensor.gi
@@ -12,6 +12,7 @@
 ##  decompositions of matrix groups.
 ##
 #############################################################################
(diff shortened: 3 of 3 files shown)


fingolfin:
Yes, that seems plausible. Although do not dare confirm your suspicions on `d247.gi` having replaced two of those files right now; last activity on each of them was close by in the SVN log. Perhaps @neunhoef remembers?

fingolfin:
Actually, code from `tensor.gi` *is* hooked up: `FindHomMethodsProjective.TensorDecomposable` is in the methods database. But perhaps you are saying that one can prove it never gets called because some earlier method gobbles up any input it might otherwise accept?

ssiccha:
@fingolfin Our guess was that Max and Akos were working on algorithms that would handle cases 2, 4 and 7. Maybe some cases are not handled by `d247.gi` (yet)?

We could try to recognize a tensor decomposable group and look at the code coverage.

fingolfin:
I just run through the examples bundled with `recog`, in `tst`. They all ended up using D247_D4TensorDec instead of `tensor`, so I think you are right.

Now I wonder about `FindHomMethodsProjective.FindElmOfEvenNormal` -- did you look at that? Its comment says: "find D2, D4 or D7 by finding an element of an even normal subgroup", and it is only called after `FindHomMethodsProjective.D247`, so I wonder if it obsolete, too?

ssiccha:
IIRC Max and Akos worked on the `RECOG.InvolutionJumper` method (but it is not finished?). So I think the question is in which situations does the involution jumper not work, and thus the other methods would be called.

neunhoef:
Hi Sergio,

the cases D2, D4 and D7 are far from being completely handled in
d247.gi. As far as I know there is still no completely analysed
algorithm with good complexity available. Also note that the classes
D2, D4 and D7 are subtly different from the usual Aschbacher classes C2,
C4 and C7. I have defined the D classes in my habilitation thesis and
the idea of the change was to make it more accessible to analysed
algorithms with good complexity bounds than the original Aschbacher
classes. This has so far failed.

Cheers,
  Max.

fingolfin:
It turns out these files *are* used after all. This can easily be seen on https://codecov.io/gh/gap-packages/recog/tree/master/gap now, but also issue #33 involves code in tensor.gi.

Hence, I removed these comments again.

PR #28, opened 2017-12-12
Add example for recognizing An and Sn on non deleted permutation modules

ssiccha:
Currently, if e.g. an Sn is given acting on a module that is not a
deleted permutation module, recognition is very inefficient.
An example is given in the file examples/NonDeletedPermModuleRepOfSn.g

### DIFF
--- a/tst/slow/NonDeletedPermModuleRepOfSn.tst
+++ b/tst/slow/NonDeletedPermModuleRepOfSn.tst
@@ -1,0 +1,13 @@
+gap> SetInfoLevel(InfoRecog, 0);;
+gap> gens := GeneratorsOfGroup;;
+gap> l := gens(SymmetricGroup(10));;
+gap> ll := List(l,x->PermutationMat(x,10)) * Z(5)^0;;
+gap> m := GModuleByMats(ll,GF(5));;
+gap> s5 := List(ll,x->KroneckerProduct(x,x));;
+gap> m := GModuleByMats(s5,GF(5));;
+gap> c := MTX.CompositionFactors(m);;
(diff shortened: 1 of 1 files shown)


gap/almostsimple.gi
```diff
@@ -930,6 +930,15 @@ FindHomMethodsProjective.AltSymBBByDegree := function(ri,G)
...
   fi;
+# FIXME: This code should go. Instead the degree of the permutation should
+# be determined by using Jambor et al.
+# For the source see:
+# https://github.com/aniemeyer/recogalmostsimple/
+# in the subfolder to-add/AnSn
+# For the paper:
+# Jambor, Sebastian, Leuner, Martin, Niemeyer, Alice C., PLESKEN, W.
+# Fast recognition of alternating groups of unknown degree.
+# J. Algebra 392 (2013), 315-335.
```
fingolfin: Hmm, wouldn't this make more sense as an issue?

ssiccha:
See #29.

examples/NonDeletedPermModuleRepOfSn.g
```diff
@@ -0,0 +1,18 @@
+gens:=GeneratorsOfGroup;
+l:=gens(SymmetricGroup(30));
+ll:=List(l,x->PermutationMat(x,30));
+g:=Group(ll*Z(5)^0);
+SetInfoLevel(InfoRecog,3);
+RecognizeGroup(g);
+m:=GModuleByMats(ll*Z(5)^0);
```
fingolfin: This line gives a syntax error
```suggestion
```

PR #43, opened 2019-03-20
Remove blackbox code

fingolfin:
This code did nothing. It is also clear in which kind of situation
it would ever be called.

In any case, we can easily add it back should we ever come into a
situation where it seems useful, but for now, it just distracts
people who try to learn how recog works.

### DIFF
--- a/doc/intro.xml
+++ b/doc/intro.xml
@@ -22,7 +22,7 @@ This package is about group recognition. It provides a generic
 framework to implement methods of group recognition, regardless of
 what computational representation is used. This means, that the code
 in this package is useful at least for permutation groups, matrix
-groups, projective groups and for the context of black box groups.
+groups and projective groups.
 The setup is described in <Cite Key="IssacRecog"/>. <P/>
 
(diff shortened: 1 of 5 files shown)


fingolfin:
@hulpke @aniemeyer any last remarks on this?

fingolfin:
@aniemeyer asked where blackbox method would then be hooked up. The answer is: in the homomorphism "databases" `FindHomDbMatrix` ,`FindHomDbPerm` and `FindHomDbPerm`, as is the case right now. While in theory it seems useful to have a database with black box methods, in practice it is now, because we need to rank all methods globally anyway. And it is not necessarily true that one should run e.g. all matrix group specific methods before any BB methods.

DominikBernhardt:
> In the homomorphism "databases" FindHomDbMatrix ,FindHomDbPerm and FindHomDbPerm, as is the case right now.

I started the integration of the new SnAnBB code in that fashion already.

hulpke:
Fine with me. Thanks for asking.

PR #63, opened 2019-07-01
Fix typos in PackageInfo.g

ssiccha:


PackageInfo.g
```diff
@@ -233,8 +233,8 @@ ArchiveFormats := ".tar.gz .tar.bz2",
 ##
 AbstractHTML :=
   "<b>Warning:</b> This package is still under development and \
-   this version is to be considered a working, but preliminary one. <p/> \
-   This packages contains a collection of methods for the \
+   this version is to be considered a working, but preliminary one. </p> \
```
wilfwilson: I don't think you can use a `P` tag on its own in HTML. GAPDoc lets you do a `<P/>` tag to make a paragraph break, which is what was here before, but that's not HTML. `P` tags in HTML have to wrap the text that is the paragraph they define, eg. `<p>This is a paragraph.</p>`. Alternatively, I think you could use a `<br />` tag, which is simply a line break.

ssiccha: Thanks. I made it a `<br/>` tag then.

ssiccha: Actually, I think a `<p>..</p>` makes more sense here.

wilfwilson: It probably makes sense to have `p` tags around both paragraphs then :)

ssiccha: :+1:

PR #67, opened 2019-07-02
Throw an error when making the documentation but pkg is already loaded

### LINKED ISSUE #64: Documentation: gap makedoc.g throws error when package is loaded automatically.
If the package is loaded automatically by gap, one cannot run makedoc.g with gap.

```
Error, another version of package recog is already loaded at /home/friedrich/gap-4.10.1/lib/package.gi:1678 called from <function "SetPackagePath">( <arguments> )
 called from read-eval loop at regen_doc.g:3
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
```

ssiccha:
Now the `regen_doc.g` file tells the user to start GAP with the `-A`
command line argument.

Fixes #64.

### DIFF
--- a/regen_doc.g
+++ b/regen_doc.g
@@ -1,6 +1,4 @@
 # Regenerate parts of the recog documentation from its source code
-
-SetPackagePath("recog", "./");
 LoadPackage("recog");
 
 GenerateMethodsXML := function(shortname, desc, db)


regen_doc.g
```diff
@@ -1,5 +1,9 @@
 # Regenerate parts of the recog documentation from its source code
 
+if IsPackageLoaded("recog") then
+    ErrorNoReturn("another version of package recog is already loaded. ",
```
DominikBernhardt: of the package recog instead of package recog

fingolfin:
I am not really that keen on merging this. See also my remark on issue #64

regen_doc.g
ssiccha: Thanks.

ssiccha:
Updated due to the discussion in #64.

ssiccha:
Rebased onto master

regen_doc.g
ssiccha: I've changed the commit and removed this if block. :)

regen_doc.g
```diff
@@ -1,6 +1,4 @@
 # Regenerate parts of the recog documentation from its source code
-
-SetPackagePath("recog", "./");
```
fingolfin: The problem with that is that we can easily end up loading the "wrong" recog, and then we also generate writing wrong documentation.

ssiccha: Ah, I thought it was not that easy to load the wrong recog. I inferred from your other comments that AutoDoc would complain if an older version was loaded. Thus I thought this could only be problematic for people that have two different versions of say the development version. And I hope this situating does not occur when compiling the documentation of a release. Or am I wrong here? 

I usually use the `makedoc.g` to compile the documentation of any manual. So I thought it would be a good idea to tell people that they should call `gap -A makedoc.g`. The call `gap makedoc.gʼ will not succeed unless one removes the recog release from the `GAPROOT/pkg` directory.

ssiccha: I somehow missed your comment in #64. What you said here sounds good to me:
> What could be done, though, is this: if recog is not loaded, use SetPackagePath to load the correct version. If it is loaded, inspect its package info record, and check its InstallationPath to verify that it refers to the correct directory.

ssiccha: Is there some way to get the absolute path that a directory object, e.g.  `Directory("./")`, represents?

fingolfin: You can use `IO_getcwd` from the IO package to get the path of the current directory; this was added in IO version 4.5.4 (2018-09-19) specifically for that purpose.

wilfwilson:
Just to note that the changes in this PR no longer seem to correspond to the title or description of this PR.

PR #76, opened 2019-07-02
Add options and improve printing of tests

ChrisJefferson:
This makes tests print even more information, and also adds some configurable options.

The main new feature, testing elements not in the group, is disabled by default as it fails for almost everything. A cycle perm test is added just to test this option works.

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -970,9 +970,40 @@ InstallGlobalFunction( "GetCompositionTreeNode",
 
 # Testing:
 
-RECOG.TestGroup := function(g,proj,size)
-  local l,r,ri,s,x,count,lvl,seedMT,seedRS,gens;
+RECOG.TestGroupOptions := rec(
+      # Number of times to test whether the recognized size is right
+      sizeTests := 3,
+
+      # Number of random elements in group to check
+      # This is used both for the number of elements in the
+      # group to check, and the number of random elements of
+      # a supergroup to check
+      inTests := 30,
+
+      # The following options are off by default as they fail on
+      # many examples at present
+
+      # if the following is set to true, then we test what happens if  SLPforElement
+      # is called with elements outside the group
+      tryNonGroupElements := false
+  );
+
+
+RECOG.TestGroup := function(g,proj,size, optionlist...)
+  local l,r,ri,s,x,count,lvl,seedMT,seedRS,gens,supergroup, options;
   count := 0;
+  
+  options := ShallowCopy(RECOG.TestGroupOptions);
+
+  if Length(optionlist) > 0 then
+    for r in RecNames(optionlist[1]) do
+        if not IsBound(options.(r)) then
+            ErrorNoReturn("Invalid option to TestGroup: ", r);
+        fi;
+        options.(r) := optionlist[1].(r);
+    od;
+  fi;
+
   lvl:=InfoLevel(InfoRecog);
   SetInfoLevel(InfoRecog, 0);
   repeat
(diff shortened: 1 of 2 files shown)


gap/base/recognition.gi
```diff
@@ -970,9 +970,37 @@ InstallGlobalFunction( "GetCompositionTreeNode",
 
 # Testing:
 
-RECOG.TestGroup := function(g,proj,size)
-  local l,r,ri,s,x,count,lvl,seedMT,seedRS,gens;
+RECOG.TestGroupOptions := rec(
+      # Number of times to test size of RecogniseGroup is right
```
fingolfin: I had to read that several times to figure out what it was saying. Perhaps this is better?
```suggestion
      # Number of times to test whether the recognized size is right
```

gap/base/recognition.gi
```diff
@@ -970,9 +970,37 @@ InstallGlobalFunction( "GetCompositionTreeNode",
...
+
+      # Number of random elements in group to check
+      # This is used both for the number of elements in the
+      # group to check, and the number of random elements of
+      # a supergroup to check
+      inTests := 30,
+
+      # The following options are off by default as they fail on
+      # many examples at present
+      tryNonGroupElements := false
```
fingolfin: Shouldn't there be an empty line between the preceeding comment and this, and then perhaps a new comment inserted which explains what `tryNonGroupElements` is about? Like this:
```suggestion

      # if the following is set to true, then we test what happens if  SLPforElement
      # is called with elements outside the group
      tryNonGroupElements := false
```

gap/base/recognition.gi
```diff
@@ -1017,10 +1048,52 @@ RECOG.TestGroup := function(g,proj,size)
...
           Error("Alarm: SLPforElement did not work!\n");
+          
           if count = -1 then return fail; fi;
       fi;
-  until count >= 30;
+  until count >= options.inTests;
+
+  if IsPermGroup(g) then
+    supergroup := SymmetricGroup(LargestMovedPoint(g) + 2);
+  elif IsMatrixGroup(g) then
```
ssiccha: Should there be a small comment here as to why we don't need `PGL(...) ` as a supergroup? 

If I understand this correctly this works because both a node for non-projective groups and a node for projective groups accept plain matrices (e.g. an `Is8BitRepMatrix`) .

ssiccha:
I like it. :)

gap/base/recognition.gi
ChrisJefferson: Longer term we should discuss what the 'supergroup' here should be. As Frank demonstrated, in GAP you can normally ask questions like `"cheese" in SymmetricGroup(3)` (which returns false), so maybe that should also be allowed here (at the moment it breaks, so I thought I'd start by only checking "sensible" things).

ssiccha: Given an inner recognition node with a reduction  homomorphism I think there always is a maximal subgroup of (P)GL that admits this homonorphism. We could generate random elements from this group and send them through the subtree rooted in the current node. 

E.g. if our group is reducible, then after a base change it lies in a group of block triangular matrices. The corresponding reduction homomorphism is a projection onto one of the diagonal blocks.

ssiccha: Can we already access this maximal subgroup from a recognition node? If not this could be one further issue.

fingolfin: I don't understand what you mean with "this maximal subgroup"; in particular, when wouldn't it be all of GL resp. PGL? Anyway, I don't think this is relevant for merging this PR, is it?

ssiccha: Oh, it's not relevant for this PR.  I'll have to open a separate issue for this and write the following thoughts down cleanly:

I was thinking about a way how to construct in our setting an overgroup of `g` which is "close" to `g`, as @ChrisJefferson put it. If we take a random elements of GL we test whether a reduction homomorphism correctly decides whether it can be applied to the random element. I think we (almost) never test the case where we generate a random element that we can apply the homomorphism to but which is not an element of `g`.

ChrisJefferson: I would be happy to see a bunch of extensions to this which generate "interesting" elements (either in, or not in, the group) which we think are likely to cause problems.

PR #79, opened 2019-07-02
Restructure test directories

MWhybrow92:
Restructure test directories so that tests are organised into those that are currently working and those that are currently broken.


fingolfin:
Unfortunately there is a conflict in `tst/quick/bugfix.tst` -- could you perhaps resolve that?

ssiccha:
Looks good to me too. :+1:

PR #80, opened 2019-07-02
Add two entries to the bibliography

wilfwilson:
These are the works cited at the top of `gap/almostsimple/lietype.gi`. I have replaced that text with the key for these works in the bibliography.

Since these new references are not actually cited yet in the documentation, they don't appear in the bibliography of the documentation when it is compiled. Is there a way of forcing all entries to appear in the bibliography, even though they're not cited (yet)? I'm not yet sure where the appropriate place would be in the manual to cite them yet.

_(This PR also fixes a tiny typo that I found, in a separate commit.)_

### DIFF
--- a/doc/recog.bib
+++ b/doc/recog.bib
@@ -12,3 +12,33 @@
        DOI = {10.1145/1145768.1145811},
       NOTE = {https://doi.org/10.1145/1145768.1145811},
 }
+
+@article {BKPS02,
+  AUTHOR = {Babai, László and Kantor, William M. and Pálfy,
+            Péter P. and Seress, Ákos}
+   TITLE = {Black-box recognition of finite simple groups of Lie type by
+            statistics of element orders},
+ JOURNAL = {J. Group Theory},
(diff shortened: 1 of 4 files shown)


doc/recog.bib
```diff
@@ -9,4 +9,35 @@ @incollection {IssacRecog
...
+FJOURNAL = {Journal of Group Theory},
+    YEAR = {2002},
+    ISSN = {1433-5883},
+ MRCLASS = {20D06 (20P05)},
+MRNUMBER = {1931364},
+     DOI = {10.1515/jgth.2002.010},
+    NOTE = {https://doi.org/10.1515/jgth.2002.010},
+}
+
+@incollection {AltseimerBorovik2001,
```
fingolfin: In the `.gi` file, you are using `AB01` instead. Maybe those should be made to match?

wilfwilson:
@fingolfin Thanks for the suggestions, I've incorporated them.

Issue #84, opened 2019-07-03
SLPforElement returns incorrect values in the trivialgroup for matrix and permutation groups


ChrisJefferson:
SLPforElement always returns a straightlineprogram for the trivial group. It should fail if given something which is not the identity.

I have a fix for this, just waiting for some other testing-related merges to happen first.

Issue #83, opened 2019-07-03
SetStabChainImmutable does nothing when a group already has a StabChainImmutable


ChrisJefferson:
In a few places in perm.gi, we call `SetStabChainImmutable` to attach a particular stabiliser chain to a group. If that group already has a stabilizer chain (which can occur during recog), then this call does nothing.

If this happens, future calls to SLPforElement produce the wrong answer, as recog expects the immutable stabilizer chain to be the one which was set.

I suspect the answer is to just keep the stabilizer chain in the recog structure, instead of trying to reattach it to the group.

PR #87, opened 2019-07-03
Various fixes

ChrisJefferson:
* Make SLPforElement work on the identity group.
* Move stabilizerchains into the structure tree instead of stored in the group (where there might be already be a different stabilizer chain).
* Add tests for these things
* Add a general test function to test all subgroups of a group, and use.
* Add a function to test all subgroups of an atlas group (not yet called)
* Move a test, which ended up in the wrong place after some merges.

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -980,6 +980,10 @@ RECOG.TestGroupOptions := rec(
   );
 
 
+# Test recognition of a group 'g' with known size 'size'.
+# 'proj' denotes if the group is projective
+# 'optionlist' is an optional list of options overriding
+# RECOG.TestGroupOptions
 RECOG.TestGroup := function(g,proj,size, optionlist...)
   local l,r,ri,s,x,count,lvl,seedMT,seedRS,gens,supergroup, options;
   count := 0;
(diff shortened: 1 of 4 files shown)


ChrisJefferson:
Fixes #84 
Fixes #83

ChrisJefferson:
EDIT: Removed handling projective groups, just to get tests to pass

fingolfin:
Annoying test failure due to RNG fluctuations:
```
# Input is:
TestSporadic("HS.2");
# Expected output:
[ "HS.2" ]
# But found:
[ "McL.2" ]
```

For now you can resolve this by commenting out that test with a comment, like several others in that file.

gap/matrix.gi
```diff
@@ -14,7 +14,11 @@
 
 SLPforElementFuncsMatrix.TrivialMatrixGroup :=
    function(ri,g)
-     return StraightLineProgramNC( [ [1,0] ], 1 );
+     if(ri!.isone(g)) then
```
fingolfin:
```suggestion
     if ri!.isone(g) then
```

gap/base/recognition.gi
```diff
@@ -1085,6 +1089,53 @@ RECOG.TestGroup := function(g,proj,size, optionlist...)
...
+# under all known representations which are permutation groups or
+# matrices over finite fields.
+# Optionally give 'options' to pass options to RECOG.TestGroup
+RECOG.testAllMaximalSubgroupsOfAtlasGroup := function(name, options...)
+    local reps, rep, maximal, exaustList;
+
+    # The following function makes it easy to extract
+    # all results of a function which returns 'fail'
+    # when given too large a value
+    exaustList := function(F)
```
```suggestion
    exhaustList := function(F)
```
and similar elsewhere :-)

gap/perm.gi
```diff
@@ -14,7 +14,11 @@
 
 SLPforElementFuncsPerm.TrivialPermGroup :=
    function(ri,g)
-     return StraightLineProgram( [ [1,0] ], 1 );
+     if(ri!.isone(g)) then
```
fingolfin:
```suggestion
     if ri!.isone(g) then
```

gap/perm.gi
```diff
@@ -270,12 +274,11 @@ end;
 SLPforElementFuncsPerm.StabChain :=
   function( ri, g )
     # we know that g is an element of Grp(ri) all without memory.
-    # we know that Grp(ri) has an immutable StabChain and
+    # we know that ri!.stabilizer is an immutable StabChain and
```
fingolfin:
```suggestion
    # we know that ri!.stabilizerchain is an immutable StabChain and
```

gap/projective.gi
```diff
@@ -14,7 +14,7 @@
 
 SLPforElementFuncsProjective.TrivialProjectiveGroup :=
    function(ri,g)
-     return StraightLineProgramNC( [ [1,0] ], 1 );
+     return StraightLineProgram( [ [1,0] ], 1 );
```
fingolfin: Why? I mean, OK, it probably doesn't matter either way -- but here we really know the SLP is correct, no needs for checks (that said, I have no idea what checks the NC version omits)

ChrisJefferson: Sorry, cut and paste error.

fingolfin:
I took the liberty to rebase this and fix a few of the things I pointed out. Note that the fix for trivial groups clashes with PR #96, but that should be trivial to resolve (and has the advantage that now only one spot has to be changed instead of three).

ChrisJefferson:
Not I'm not currently fixing trivial projective groups, as doing so breaks a bunch of tests (and I haven't investigated why yet)

fingolfin:
The failures in the test for trivial projective group are real, serious bugs. Some of the tests simply have wrong validation sizes. Ouch.

ChrisJefferson:
I've dropped the trivialgroup part, which is now handled much more completely in #102

PR #98, opened 2019-07-03
Add ValidateHomomInput, and use it to fix ThrowAwayFixedPoints

ChrisJefferson:
This adds a new attribute, HomomSourceCheck, which checks if a value is in the source of Homom.

This is needed (I believe) to let us reject invalid elements of groups, but I'm open to discussion.

(EDIT: was called Range, that was the wrong name).

### DIFF
--- a/doc/recognition.xml
+++ b/doc/recognition.xml
@@ -249,6 +249,15 @@ kernel have been created or recognised yet. You can use for example
 the methods in <Ref Var="FindHomMethodsPerm"/> or
 <Ref Var="FindHomMethodsMatrix"/> or <Ref Var="FindHomMethodsProjective"/>
 as the <A>method</A> argument.
+
+<P/>GAP homomorphisms are not required to give a sensible answer when
+given a value not in their source, and in practice often enter the break
+loop, or return an incorrect answer. This causes problems when checking
+if a value is not in the represented group. To avoid this problem,
+<Ref Attr="validatehomominput"/> can be set to a function.
+This function is used to filter possible group elements,
+before they are
+passed to <Ref Attr="Homom"/>.
 </Description>
 </ManSection>
 
@@ -295,6 +304,45 @@ The following attributes are defined for recognition info records:
 <#Include Label="SLPforElement">
 <#Include Label="StdPresentation">
 <#Include Label="methodsforfactor">
+
+<ManSection>
+<Attr Name="validatehomominput" Arg="ri, x"/>
+<Description>
+
+The value of this attribute, if there is any, must be a function with
+two arguments: a recognition record <A>ri</A>, and an element <A>x</A>.
+The function must return a boolean. If it returns <K>false</K>, then
+this means that <A>x</A> is not in the source of the homomorphism
+returned by <Ref Attr="Homom"/>. If <K>true</K> is returned, then
+either <A>x</A> is in the source of that homomorphism, or passing
+<A>x</A> to the homomorphism returns <K>fail</K>.</P>
+
+For example,
+if <A>ri</A> represents a matrix group that preserves a subspace, then
+the source of <Ref Attr="Homom"/> will be matrices which preserve that
+subspace, and passing matrices which do not preserve this subspace
+to <Ref Attr="Homom"/> may produce incorrect answers.
+<Ref Attr="validatehomominput"/> can be used to filter out such elements.
+
+The function <Ref Func="ValidateHomomInput"/> provides a simple wrapper
+to this attribute, which calls <Ref Attr="validatehomominput"/> unless
+it is not defined, in which case <Ref Func="ValidateHomomInput"/>
+returns true.
+</Description>
+</ManSection>
+
+
+<ManSection>
+<Func Name="ValidateHomomInput" Arg="ri, x"/>
+<Returns>a boolean.
+</Returns>
+<Description>
+    This is a wrapper function which calls <Ref Attr="validatehomominput"/>
+    of <A>ri</A> with <A>x</A>, or returns true if <A>ri</A> does not define
+    <Ref Attr="validatehomominput"/>.
+</Description>
+</ManSection>
+
 
 The following two attributes are concerned with the relation between
 the original generators and the nice generators for a node.
@@ -657,7 +705,10 @@ A successful find homomorphism method has the following duties:
     group in the attribute <Ref Attr="Homom"/>. Note that if your homomorphism
     changes the representation (for example going from matrix groups to
     permutation groups), you will have to set the attribute
-    <Ref Attr="methodsforfactor"/> accordingly.
+    <Ref Attr="methodsforfactor"/> accordingly. Also,
+    <Ref Attr="ValidateHomomInput"/> may be set to a function which returns
+    <C>false</C> for values which may cause <Ref Attr="Homom"/> to
+    produce the wrong answer, or error.
 </Item>
 <Mark>for non-leaves: kernel generators</Mark>
 <Item>
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -528,6 +528,7 @@ InstallGlobalFunction( RecogniseGeneric,
               counter,").");
         fi;
         Add(depth,'F');
+        Assert(2, ForAll(GeneratorsOfGroup(H), x->ValidateHomomInput(ri, x)));
         rifac := RecogniseGeneric(
                   Group(List(GeneratorsOfGroup(H), x->ImageElm(Homom(ri),x))),
                   methodsforfactor(ri), depth, forfactor(ri) ); # TODO: change forfactor to hintsForFactor??)
@@ -628,6 +629,7 @@ InstallGlobalFunction( RecogniseGeneric,
                 # We must use different random elements than the kernel
                 # finding routines!
                 x := RandomElm(ri,"KERNELANDVERIFY",true).el;
+                Assert(2, ValidateHomomInput(ri, x));
                 s := SLPforElement(rifac,ImageElm( Homom(ri), x!.el ));
                 if s = fail then
                     ErrorNoReturn("Very bad: factor was wrongly recognised and we ",
@@ -674,6 +676,15 @@ InstallGlobalFunction( RecogniseGeneric,
     # StopStoringRandEls(ri);
     return ri;
   end);
+
+InstallGlobalFunction( ValidateHomomInput,
+  function(ri,x)
+    if Hasvalidatehomominput(ri) then
+        return validatehomominput(ri)(ri,x);
+    else
+        return true;
+    fi;
+  end );
 
 InstallGlobalFunction( CalcNiceGens,
   function(ri,origgens)
@@ -723,6 +734,11 @@ InstallGlobalFunction( SLPforElementGeneric,
     local gg,n,rifac,riker,s,s1,s2,y,nr1,nr2;
     rifac := RIFac(ri);
     riker := RIKer(ri);   # note: might be fail
+
+    if not ValidateHomomInput(ri, g) then
+        return fail;
+    fi;
+
     gg := ImageElm(Homom(ri),g);
     if gg = fail then
         return fail;
@@ -764,6 +780,7 @@ InstallGlobalFunction( FindKernelRandom,
     rifac := RIFac(ri);
     for i in [1..n] do
         x := RandomElm(ri,"KERNELANDVERIFY",true).el;
+        Assert(2, ValidateHomomInput(ri, x));
         s := SLPforElement(rifac,ImageElm( Homom(ri), x!.el ));
         if s = fail then
             return false;
--- a/gap/base/recognition.gd
+++ b/gap/base/recognition.gd
@@ -195,6 +195,7 @@ InstallTrueMethod( IsRecogInfoForAlmostSimpleGroup, IsRecogInfoForSimpleGroup );
 ## </ManSection>
 ## <#/GAPDoc>
 DeclareAttribute( "pregensfac", IsRecognitionInfo, "mutable" );
+DeclareAttribute( "validatehomominput", IsRecognitionInfo);
 
 ## <#GAPDoc Label="calcnicegens">
 ## <ManSection>
@@ -500,6 +501,7 @@ DeclareGlobalFunction( "TryFindHomMethod" );
 ## </ManSection>
 ## <#/GAPDoc>
 DeclareGlobalFunction( "CalcNiceGens" );
+DeclareGlobalFunction( "ValidateHomomInput" );
 
 ## <#GAPDoc Label="CalcNiceGensGeneric">
 ## <ManSection>
--- a/gap/matrix.gi
+++ b/gap/matrix.gi
@@ -689,6 +689,7 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
     rifac := RIFac(ri);
     while nothingnew < 10 do
         x := RandomElm(ri,"KERNEL",true).el;
+        Assert(2, ValidateHomomInput(ri, x));
         s := SLPforElement(rifac,ImageElm( Homom(ri), x!.el ));
         y := ResultOfStraightLineProgram(s,
                  ri!.genswithmem{[ri!.nrgensH+1..Length(ri!.genswithmem)]});
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -362,6 +362,7 @@ FindHomMethodsPerm.ThrowAwayFixedPoints :=
       o := MovedPoints(G);
       hom := ActionHomomorphism(G,o);
       SetHomom(ri,hom);
+      Setvalidatehomominput(ri, {ri,p} -> IsSubset(o, MovedPoints(p)));
 
       # Initialize the rest of the record:
       findgensNmeth(ri).method := FindKernelDoNothing;
--- a/tst/broken/veryslow/ClassicalNatural.tst
+++ b/tst/broken/veryslow/ClassicalNatural.tst
@@ -23,3 +23,9 @@ gap> TestRecogGL(9,5);; # FIXME buggy, see issue #37
 gap> TestRecogGL(2,8);; # FIXME: see issue #12
 gap> TestRecogGL(2,16);; # FIXME: see issue #12
 gap> TestRecogGL(8,27);; # FIXME: see issue #12
+
+gap> TestRecogGL(10,3);; # FIXME
+gap> TestRecogGL(18,3);; # FIXME
+gap> TestRecogGL(3,27);; # FIXME
+
+
--- a/tst/working/quick/PermThrowAwayFixedPoints.txt
+++ b/tst/working/quick/PermThrowAwayFixedPoints.txt
@@ -1,0 +1,7 @@
+gap> g := Group( (100,101,102), (101,102) );;
+gap> RECOG.TestGroup(g, false, 6, rec(tryNonGroupElements := true));;
+
+# Explicitly test failing case
+gap> ri := RecogniseGroup(g);;
+gap> SLPforElement(ri, (1,2));
+fail
--- a/tst/working/slow/basic.tst
+++ b/tst/working/slow/basic.tst
@@ -40,13 +40,13 @@ gap> ri := RECOG.TestGroup(g,false,24^32);;
 
 # Test for ThrowAwayFixedPoints:
 gap> g := Group( (100,101,102,103,104,105,106,107,108,109,110), (100,101) );;
-gap> ri := RECOG.TestGroup(g,false,Factorial(11));;
+gap> ri := RECOG.TestGroup(g,false,Factorial(11), rec(tryNonGroupElements := true));;
 
 # Test for ThrowAwayFixedPoints 2:
 gap> x := PermList(Concatenation([2..1000],[1],[1002,1003,1001]))^1000;;
 gap> y := PermList(Concatenation([2..1000],[1],[1001,1002,1004,1005,1003]))^1000;;
 gap> g := Group( x,y );;
-gap> ri := RECOG.TestGroup(g,false,60);;
+gap> ri := RECOG.TestGroup(g,false,60, rec(tryNonGroupElements := true));;
 
 #
 gap> gg := PrimitiveGroup(102,1);;
--- a/tst/working/veryslow/ClassicalNatural.tst
+++ b/tst/working/veryslow/ClassicalNatural.tst
@@ -61,11 +61,7 @@ gap> TestRecogGL(8,3);;
 Stamp: ClassicalNatural
 gap> TestRecogGL(9,3);;
 Stamp: ClassicalNatural
-gap> TestRecogGL(10,3);;
-Stamp: ClassicalNatural
 gap> TestRecogGL(17,3);;
-Stamp: ClassicalNatural
-gap> TestRecogGL(18,3);;
 Stamp: ClassicalNatural
 gap> TestRecogGL(19,3);;
 Stamp: ClassicalNatural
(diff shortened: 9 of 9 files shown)


doc/recognition.xml
```diff
@@ -416,6 +419,19 @@ elements that are generated from them remember, how they were acquired.
 </Description>
 </ManSection>
 
+<ManSection>
+<Attr Name="HomomRangeCheck" Arg="ri"/>
```
ssiccha: I think a better name is `HomomDomainCheck`. 

In the case you describe below the domain of the homomorphism consists of those matrices that leave the corresponding subspace invariant. 
The new function checks whether its argument is in the domain of the homomorphism.

doc/recognition.xml
```diff
@@ -416,6 +419,19 @@ elements that are generated from them remember, how they were acquired.
...
+<Description>
+    The value of this attribute is a function which returns a boolean
+    which is <K>false</K> for values which should not be passed to
```
ssiccha: which are not in the domain of

doc/recognition.xml
```diff
@@ -416,6 +419,19 @@ elements that are generated from them remember, how they were acquired.
...
+    <Ref Attr="Homom"/>. For example, if this node represents
+    a subspace preserved by a matrix group, then matrices which do not
```
ssiccha: a matrix group that preserves a subspace

doc/recognition.xml
ssiccha:
```suggestion
    which is <K>false</K> for values which which are not in the domain of
```

doc/recognition.xml
ssiccha:
```suggestion
    a matrix group that preserves a subspace, then matrices which do not
```

doc/recognition.xml
ChrisJefferson: my reason for using Range is (I think, I don't really use homomorphisms in GAP very much) that GAP uses Range(hom), rather than Domain(hom), so the name lines up (somewhat) with GAP. However, I'm happy to be (a) wrong, or (b) just more standard maths names.

ChrisJefferson:
Note that I think many (almost all?) uses of SetHomom probably need a SetHomomRangeCheck at some point, I just did ThrowAwayUsedPoints as a simple example

doc/recognition.xml
```diff
@@ -984,7 +1000,13 @@ A successful find homomorphism method has the following duties:
     group in the attribute <Ref Attr="Homom"/>. Note that if your homomorphism
     changes the representation (for example going from matrix groups to
     permutation groups), you will have to set the attribute
-    <Ref Attr="methodsforfactor"/> accordingly.
+    <Ref Attr="methodsforfactor"/> accordingly. Also,
+    <Ref Attr="HomomRangeCheck"/> may be set to a function which returns
+    <C>false</C> for values which may cause <Ref Attr="Homom"/> to
+    produce the wrong answer, or error. For example, if this node represents
+    a subspace preserved by a matrix group, then matrices which do not
```
ssiccha:
```suggestion
    a matrix group which preserves a subspace, then matrices which do not
```

doc/recognition.xml
ssiccha: I checked the documentation. `Source` is the name of the source and `Range` the name of the target.

ssiccha:
> Note that I think many (almost all?) uses of SetHomom probably need a SetHomomRangeCheck at some point, I just did ThrowAwayUsedPoints as a simple example

I agree.

doc/recognition.xml
ChrisJefferson: Yes, that's just my inability to do.. words.

ChrisJefferson: thanks

doc/recognition.xml
```diff
@@ -336,8 +336,11 @@ set this to <K>false</K>. The result is either <K>fail</K> or
 <K>false</K> if the method fails or a recognition info record <C>ri</C>.
 If the method created a leaf then <C>ri</C> will be a leaf, otherwise
 it will have the attribute <Ref Attr="Homom"/> set, but no factor or
-kernel have been created or recognised yet. You can use for example
-the methods in <Ref Var="FindHomMethodsPerm"/> or
+kernel have been created or recognised yet. Optionally,
+<Ref Attr="HomomSourceCheck"/> can be set. When this is set, <Ref Attr="Homom"/>
+can fail, error, or produce incorrect answers for any value for which
+<Ref Attr="HomomSourceCheck"/> returns false.
+You can use for example the methods in <Ref Var="FindHomMethodsPerm"/> or
```
fingolfin: I really think this needs more explanation as to *why* this matters. From just reading this, it is not clear to me.

Naively I would say: "why can't the Homom homomorphism just return `fail`, and be done with it? I am guessing the answer is: "because some of them are stock GAP homomorphisms, and they produces errors.". Though perhaps it is also "... and they produce nonsense results for nonsense inputs", which arguable would be worse.

That said, of course a second strategy is possible: We can just demand that morphisms must always validate their input. If we were to do that, then we wouldn't need this extra function, at the cost of making it slightly more annoying to make homomorphism objects -- though we could provide a helper function to make it very easy.

BTW, note that for `ActionHomomorphism`, one can set the `actioncanfail` option to get `fail` instead of errors -- but that requires one to pass that to every call of `ImageElm`.. Hrm.

doc/recognition.xml
```diff
@@ -416,6 +419,19 @@ elements that are generated from them remember, how they were acquired.
 </Description>
 </ManSection>
 
+<ManSection>
+<Attr Name="HomomSourceCheck" Arg="ri"/>
+<Description>
+    The value of this attribute is a function which returns a boolean
+    which is <K>false</K> for values which are not in the source of
```
fingolfin: What does "values" refer to? What values? Perhaps the function returned is supposed to be unary (?), and "values" returns to any value passed to it as an argument?

doc/recognition.xml
```diff
@@ -416,6 +419,19 @@ elements that are generated from them remember, how they were acquired.
...
+    <Ref Attr="Homom"/>. For example, if this node represents
+    a matrix group that preserves a subspace, then matrices which do not
+    preserve that subspace may cause <Ref Attr="Homom"/> to produce
+    incorrect answers. If not defined, this function defaults
```
fingolfin: OK, so it seems this is about incorrect answers, not about calls to `Error()`?

doc/recognition.xml
```diff
@@ -336,8 +336,11 @@ set this to <K>false</K>. The result is either <K>fail</K> or
 <K>false</K> if the method fails or a recognition info record <C>ri</C>.
 If the method created a leaf then <C>ri</C> will be a leaf, otherwise
 it will have the attribute <Ref Attr="Homom"/> set, but no factor or
-kernel have been created or recognised yet. You can use for example
-the methods in <Ref Var="FindHomMethodsPerm"/> or
+kernel have been created or recognised yet. Optionally,
+<Ref Attr="HomomSourceCheck"/> can be set. When this is set, <Ref Attr="Homom"/>
```
fingolfin: Hmm, I am not in love with that name. It sounds to me as if the source was checked. But rather, it is some input that is meant to be passed to the homomorphism that is checked, or validated.
So perhaps `ValidateHomomInput` or `CheckHomomInput`? Or `IsValidInputForHomom`?

doc/recognition.xml
ChrisJefferson: incorrect answers, calls to Error, break loops,... I've seen all 3. But not sure the best way of describing that.

doc/recognition.xml
ChrisJefferson: From experience this week, built in GAP homomorphisms can Error, enter the break loop, or produce incorrect answers, if given a value not in their source.

It seems like GAP's general attitude (and documentation) is that this is how homomorphisms work, so I feel like that's the more natural thing to do (and, will make it easier to disable these checks late, if/when we want). It does of course mean we need to remember to call this extra function where required.

doc/recognition.xml
ChrisJefferson: Yes, the first 'value' is the function, the second 'value' is arguments to this function (the function is unary)

doc/recognition.xml
ChrisJefferson: I like IsValidInputForHomom

ChrisJefferson: although actually, normally IsX properties in GAP return true/false, so maybe ValidateHomomInput would be better...

ChrisJefferson:
I've had a go at overhauling names and docs.

Some tests are failing, which I can move in the event we decide this is an otherwise mergable PR.

doc/recognition.xml
```diff
@@ -356,8 +356,17 @@ set this to <K>false</K>. The result is either <K>fail</K> or
 <K>false</K> if the method fails or a recognition info record <C>ri</C>.
 If the method created a leaf then <C>ri</C> will be a leaf, otherwise
 it will have the attribute <Ref Attr="Homom"/> set, but no factor or
-kernel have been created or recognised yet. You can use for example
-the methods in <Ref Var="FindHomMethodsPerm"/> or
+kernel have been created or recognised yet.
```
wilfwilson: I know it's not your sentence, but since we're touching it, could you please update to:
```suggestion
kernel will have been created or recognised yet.
```

doc/recognition.xml
```diff
@@ -356,8 +356,17 @@ set this to <K>false</K>. The result is either <K>fail</K> or
...
+
+<P/>GAP homomorphisms are not required to give a sensible answer when
+given a value not in their source, and in practice often enter the break
+loop, or return an incorrect answer. This causes problems when checking
+if a value is not in the represented group. To avoid this problem,
+<Ref Attr="ValidateHomomInput"/> can be assigned to a function.
+This function is used to filter possible group elements <C>x</C>,
```
wilfwilson: Please remove `<C>x</C>`, there's no point to define a variable that's never used 🙂

doc/recognition.xml
```diff
@@ -436,6 +445,19 @@ elements that are generated from them remember, how they were acquired.
 </Description>
 </ManSection>
 
+<ManSection>
+<Attr Name="ValidateHomomInput" Arg="ri"/>
+<Description>
+    The value of this attribute is a function. This function returns
```
wilfwilson:
```suggestion
    The value of this attribute is a function of one variable. This function returns
```

doc/recognition.xml
```diff
@@ -436,6 +445,19 @@ elements that are generated from them remember, how they were acquired.
...
+    a boolean which is <K>false</K> when passed a value which is not
```
wilfwilson: Is this really what we want? This suggests that `ValidateHomomInput` is completely testing membership of the element in the source.

I had assumed that we want a looser check. My assumption is that `ValidateHomomInput` would be required return `false` if either:
1. it is not in the source and this fact is easy to compute;
2. it is not in the source, and applying the homomorphism will produce an error/incorrect result.

and return `true` otherwise (of course, it could do more work if it wanted).

This would still allow the homomorphism to be given values for which it will correctly return `fail`.
What do you think about this?

doc/recognition.xml
```diff
@@ -1004,7 +1026,13 @@ A successful find homomorphism method has the following duties:
     group in the attribute <Ref Attr="Homom"/>. Note that if your homomorphism
     changes the representation (for example going from matrix groups to
     permutation groups), you will have to set the attribute
-    <Ref Attr="methodsforfactor"/> accordingly.
+    <Ref Attr="methodsforfactor"/> accordingly. Also,
+    <Ref Attr="ValidateHomomInput"/> may be set to a function which returns
+    <C>false</C> for values which may cause <Ref Attr="Homom"/> to
```
wilfwilson: This language is more along the lines of what I am suggesting.

doc/recognition.xml
ChrisJefferson: You are right, that's far too strong.

In general, all we really care about is the Homom works. In practice there is usually (as described elsewhere) a property which this node of the tree uses to split the group (invariant subspace, block structure, ...), and we want to check that property.

doc/recognition.xml
fingolfin:
```suggestion
This function is used to filter possible group elements,
```

gap/base/recognition.gi
```diff
@@ -708,6 +708,11 @@ InstallGlobalFunction( SLPforElementGeneric,
     local gg,n,rifac,riker,s,s1,s2,y,nr1,nr2;
     rifac := RIFac(ri);
     riker := RIKer(ri);   # note: might be fail
+
+    if HasValidateHomomInput(ri) and not ValidateHomomInput(ri)(g) then
+        return fail;
+    fi;
+
     gg := ImageElm(Homom(ri),g);
```
fingolfin: There in total 5 places where we call `ImageElm`. I think for each of the 4 other places, we should either add a call to `ValidateHomomInput`, or else add a comment that explains why a call to `ValidateHomomInput` is not necessary.

doc/recognition.xml
```diff
@@ -356,8 +356,17 @@ set this to <K>false</K>. The result is either <K>fail</K> or
 <K>false</K> if the method fails or a recognition info record <C>ri</C>.
 If the method created a leaf then <C>ri</C> will be a leaf, otherwise
 it will have the attribute <Ref Attr="Homom"/> set, but no factor or
-kernel have been created or recognised yet. You can use for example
-the methods in <Ref Var="FindHomMethodsPerm"/> or
+kernel have been created or recognised yet.
+
+<P/>GAP homomorphisms are not required to give a sensible answer when
```
ssiccha: I like the idea behind this new section a lot! This is an important information that was missing in the documentation.

Though then this should be moved to another, possibly more discoverable, location. The GAPDoc comment you are adding this section to is for `TryFindHomMethod`. But I'm pretty sure `TryFindHomMethod` was intended as a debug function, namely as a way to apply a specific "FindHomMethod" to a group. It is never called anywhere.
Also see a source code comment in `gap/base/recognition.gi` added by @fingolfin in 2017:
```
# TODO: TryFindHomMethod is never called by anything
# in recog.
# Seems to be for user intervention and/or debugging?
```

doc/recognition.xml
ssiccha: To add to that: at the beginning of the recognition we can't expect to have a membership test for the group since that's precisely what we are trying to construct.

doc/recognition.xml
ssiccha: IMO a little bit better than `ValidateHomomInput` would be `ValidateInputForHomomorphism`.

fingolfin:
@ChrisJefferson I am in principle in favor of this PR, but there are a ton of unaddressed comments on it -- probably because you didn't want to waste efforts on this PR if it then isn't merged. For now, I think we should merge it: It solves an actual problem (of which several more variants exist), and even if perhaps we may want/need to revisit our approach to this later on, that's no reason to not merge this.

So, if you are willing to go ahead and address all comments that were made so far, I'll review this once more, and then I hope we can soon move forward with merging this.

ChrisJefferson:
I am going to make another pass over this PR, based on comments, just had a busy weekend, and start of week.

doc/recognition.xml
ChrisJefferson: Wait, I've thought about this some more, I think it is right (although I can't remember if I've changed it & pushed since this comment was first made)

Note that Homom usually accepts many things which are not in the group. Things not in the image of Homom should not be given to Homom as an argument, so we have to reject them.

ChrisJefferson:
I think I've now cleaned up. I've done something which I don't really like the look of, but fits with everything else, which is make `validateinputhomom` be the attribute, and write a function `ValidateInputHomom` which wraps it (I'm happy for this to be changed, when/if there is a renaming of the similar cases).

I looked at the calls to InputElm, and I believe they should always pass things which are accepted by Homom. In these places I've put an Assert, if it turns out later these are wrong they can be changed.

I'd like to define some other validateinputhomom, but I think it would be better to merge this first.

doc/recognition.xml
wilfwilson: I see your point, assuming that you mean 'source' rather than 'image'. I just wanted to be sure that we're not insisting on a group membership test.

doc/recognition.xml
fingolfin: Regarding the name: as long as I can use a global search&replace to adjust it, I am not going to spend extended amounts of time arguing about it ;-). We of course need to carefully think about all those names, but that can happen in a follow-up PR, no need to delay the fix in this PR for it.

ChrisJefferson: I think the introduction of validatehomominput and ValidateHomomInput makes the naming particularly horrible, but that's horrible consistent naming, and I'm very happy for it to be changed so we don't rely on case :)

gap/base/recognition.gi
```diff
@@ -659,6 +660,15 @@ InstallGlobalFunction( RecogniseGeneric,
     return ri;
   end);
 
+InstallGlobalFunction( ValidateHomomInput,
+  function(ri,x)
+    if Hasvalidatehomominput(ri) then
+        return validatehomominput(ri)(x);
```
fingolfin: I wonder if we shouldn't pass `ri` to the validation function, too, for more flexibility in the future. I mean, sure, if you create that validation function as a lambda on the spot each time, you don't need that. But may be more convenient (e.g. for debugging, and for more complex validation) to define a single such function in a central place, and then pass `ri` to it.
```suggestion
        return validatehomominput(ri)(ri, x);
```

doc/recognition.xml
```diff
@@ -352,8 +352,17 @@ set this to <K>false</K>. The result is either <K>fail</K> or
...
+given a value not in their source, and in practice often enter the break
+loop, or return an incorrect answer. This causes problems when checking
+if a value is not in the represented group. To avoid this problem,
+<Ref Attr="validatehomominput"/> can be assigned to a function.
```
fingolfin: This seems to be phrased the wrong way around: not `validatehomominput` is assigned to a function, but rather a function is assigned to `validatehomominput`; or `validatehomominput` is set to a function.
```suggestion
<Ref Attr="validatehomominput"/> can be set to a function.
```

fingolfin:
Random note (sorry, I'll find time for a review on one of the next train rides, today or tomorrow): I think `RECOG.HomBCToDiagonalBlock` is another case were we need verification. And we need to think about whether a few of those homs where I added verification should be changed to use this new mechanism.

ChrisJefferson:
Now fixed. THe documentation could be moved around / made neater, but I'd prefer to merge it rather than move things around and risk breaking / conflicting with something else.

doc/recognition.xml
```diff
@@ -296,6 +305,37 @@ The following attributes are defined for recognition info records:
 <#Include Label="StdPresentation">
 <#Include Label="methodsforfactor">
 
+<ManSection>
+<Attr Name="validatehomominput" Arg="ri, x"/>
+<Description>
+    The value of this attribute is a function of two variables. This function returns
+    a boolean which is <K>false</K> when <A>x</A> is not
+    in the source of the <Ref Attr="Homom"/> of <A>ri</A>. For example,
+    if <A>ri</A> represent a matrix group that preserves a subspace, then
```
fingolfin:
```suggestion
    if <A>ri</A> represents a matrix group that preserves a subspace, then
```

doc/recognition.xml
```diff
@@ -296,6 +305,37 @@ The following attributes are defined for recognition info records:
 <#Include Label="StdPresentation">
 <#Include Label="methodsforfactor">
 
+<ManSection>
+<Attr Name="validatehomominput" Arg="ri, x"/>
+<Description>
+    The value of this attribute is a function of two variables. This function returns
```
fingolfin:
```suggestion
    The value of this attribute, if there is any, must be a function of two variables. This function returns
```

doc/recognition.xml
```diff
@@ -296,6 +305,37 @@ The following attributes are defined for recognition info records:
...
+    a boolean which is <K>false</K> when <A>x</A> is not
+    in the source of the <Ref Attr="Homom"/> of <A>ri</A>. For example,
```
fingolfin: This does not mention when another boolean is returned -- presumably only `true` is meant here (`fail` is technically also a boolean).
Anyway, technically, deciding whether an element is in the "source" may impossible at this point because in order to decide this, we'd need to have fully recognized the group. But the current language seems to require that `false` *must* be returned in this case, which is just not feasible.

Instead, I would described what it is *used* for, and then only described typical reasons for when to return true and when false later on. Maybe like this:

```
The value of this attribute, if there is any, must be a function with
two arguments: a recognition record <A>ri</A>, and an element <A>x</A>.
The function must return a boolean. If it returns <K>false</K>, then
this means that <A>x</A> cannot be in the source of the homomorphism
returned by <Ref Attr="Homom"/>. If <K>true</K> is returned, then
<A>x</A> may be contained in the source of that homomorphism, and it
must be safe to pass it to that homomorphism. I.e., that homomorphism
then must not produce an error; and if <A>x</A> turns out to be not in
the source after all, it must return <K>fail</K>.
```

Then you can go on with the example of a matrix group.

doc/recognition.xml
```diff
@@ -249,6 +249,15 @@ kernel have been created or recognised yet. You can use for example
...
 <Ref Var="FindHomMethodsMatrix"/> or <Ref Var="FindHomMethodsProjective"/>
 as the <A>method</A> argument.
+
+<P/>GAP homomorphisms are not required to give a sensible answer when
+given a value not in their source, and in practice often enter the break
+loop, or return an incorrect answer. This causes problems when checking
+if a value is not in the represented group. To avoid this problem,
+<Ref Attr="validatehomominput"/> can be set to a function.
+This function is used to filter possible group elements,
+which are not known to be in the recognised group, before they are
```
fingolfin: I am not quite sure I understand this sentence correctly. I think I'd prefer if it said a bit less:
```suggestion
before they are
```

doc/recognition.xml
```diff
@@ -657,7 +697,13 @@ A successful find homomorphism method has the following duties:
...
     changes the representation (for example going from matrix groups to
     permutation groups), you will have to set the attribute
-    <Ref Attr="methodsforfactor"/> accordingly.
+    <Ref Attr="methodsforfactor"/> accordingly. Also,
+    <Ref Attr="ValidateHomomInput"/> may be set to a function which returns
+    <C>false</C> for values which may cause <Ref Attr="Homom"/> to
+    produce the wrong answer, or error. For example, if this node represents
+    a matrix group that preserves a subspace, then matrices which do not
+    preserve that subspace may cause <Ref Attr="Homom"/> to produce
+    incorrect answers.
```
fingolfin: Is it really necessary / useful to give yet another example here? Shouldn't readers of this be expected to look at the documentation for the respective attribute for further details?

gap/base/recognition.gi
```diff
@@ -528,6 +528,7 @@ InstallGlobalFunction( RecogniseGeneric,
               counter,").");
         fi;
         Add(depth,'F');
+        Assert(2, ForAll(GeneratorsOfGroup(H), x->ValidateHomomInput(ri, x)));
```
fingolfin: There is another call to `ImageElm` around line 631, which also should at the very least get such an assertion.

doc/recognition.xml
ChrisJefferson: My reading of GAP's documentation is that the things in 'source' are exactly the values which it is, in general, legal to pass to a homomorphism.

It's not required that things in the image are in the group we are identifying, the image of the homomorphism will be in general anything which satisfies the "property" of this node (has a block system, fixes some points, whatever).

However, it does seem this point is a point of confusion.

ChrisJefferson:
I think the issues are now all resolved (although I'm sure the docs could still be improved)

Issue #101, opened 2019-07-03
Bug in recognition for lower triangular matrix groups


fingolfin:
The `quick/MatReducible.tst` test fails. It involves upper triangular 10x10 matrices over GF(5), but something goes wrong (possibly in the `FindHomMethodsMatrix.BlockScalar`). This was revealed by the addition of a validation check in the `SLPforElementFuncsGeneric.TrivialGroup` by @ChrisJefferson.

Further related broken tests are in
- `quick/ProjLowIndex.tst`
- `quick/ProjReducible.tst`
- `slow/ProjC6.tst`
- `veryslow/MatC6.tst`

For now, I will disable these tests by moving them to the `tst/broken` directory.

PR #102, opened 2019-07-03
Disable some secretely broken tests, and add validation check to SLPforElementFunc for trivial groups

### LINKED ISSUE #84: SLPforElement returns incorrect values in the trivialgroup for matrix and permutation groups

### LINKED ISSUE #91: SLPforElement produces wrong answers for the trivial projective group
```
gap> g := GroupWithGenerators([One(GL(7,5))]);;
gap> gens := ShallowCopy(GeneratorsOfGroup(g));;
gap> repeat r := Random(GF(5)); until not(IsZero(r));
gap> Add(gens,gens[1]*r);;
gap> ri := RECOG.TestGroup(g,true,1,rec(tryNonGroupElements := true));;
```

fails, as `SLPforElement(g, x)` passes for every `x`.

I have pulled this out of #84, as I fixed the perm and matrix cases but my fix for this seemed broken, which might mean I don't understand projective groups.

fingolfin:
See also issue #101; the second commit is taken from PR #87 (albeit in modified form) by @ChrisJefferson 

Fixes #84
Fixes #91

### DIFF
--- a/gap/matrix.gi
+++ b/gap/matrix.gi
@@ -667,6 +667,8 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
   function(ri)
     local b,curlay,done,el,f,i,l,lens,lvec,nothingnew,pivots,pos,ready,
           rifac,s,v,x,y;
+
+    Info(InfoRecog, 2, "Running FindKernelLowerLeftPGroup...");
     f := ri!.field;
     if not(IsPrimeField(f)) then
         b := CanonicalBasis(f);  # a basis over the prime field
--- a/gap/generic/TrivialGroup.gi
+++ b/gap/generic/TrivialGroup.gi
@@ -14,7 +14,10 @@
 #############################################################################
 
 SLPforElementFuncsGeneric.TrivialGroup := function(ri,g)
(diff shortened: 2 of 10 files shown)


ChrisJefferson:
Looks good, modulo a couple more failing tests

gap/matrix.gi
```diff
@@ -667,6 +667,8 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
   function(ri)
     local b,curlay,done,el,f,i,l,lens,lvec,nothingnew,pivots,pos,ready,
           rifac,s,v,x,y;
+
+    Info(InfoRecog, 2, "Running FindKernelLowerLeftPGroup...");
```
ssiccha: Is this here on purpose? Just wondering since it's not mentioned in the commit msg.

fingolfin: It's in there by accident, but I did mean to add it to recog eventually. It may not be "pure", but I'd just merge it along now, no point in wasting time just to get that one line into a separate commit / separate PR... :-)

PR #105, opened 2019-07-03
Fix BlocksModScalars if all blocks have size 1

### LINKED ISSUE #101: Bug in recognition for lower triangular matrix groups

fingolfin:
In that case, we are looking at a bunch of diagonal matrices as generators, which produce a trivial image. We used SLPforElementFuncsGeneric.TrivialGroup for that; however, that now checks if its input satisfies ri!.isone; but that is of course not the case here, as the input consists of arbitrary diagonal matrices. So, we need to make a variant SLPforElementFuncsGeneric.TrivialGroup which has a different input check.

Fixes #101


ChrisJefferson:
this looks right. Good catch.

Issue #78, opened 2019-07-02
Extend the bibliography


fingolfin:
E.g including these:

Max Neunhöffer, “[Constructive Recognition of Finite Groups](https://github.com/neunhoef/habil/raw/master/habil_accepted.pdf)”, 2009, habilitation thesis ([source code](https://github.com/neunhoef/habil/raw)) – more in-depth information on much of the mathematics behind recog

Henrik Bäärnhielm, Derek Holt, Charles R Leedham-Green, and Eamonn A O’Brien, [A practical model for computation with matrix groups](https://www.sciencedirect.com/science/article/pii/S074771711400056X), 2014 – survey of matrix group recognition as implemented in Magma, with an extensive bibliography

We also should generally add all papers to the bibliography from which something is implemented.

PR #106, opened 2019-07-04
Extend the bibliography

wilfwilson:
I intend for this to address, at least to some degree, #78.

It seems that GAPDoc chooses its own keys for the bib entries when compiling the documentation (e.g. `JLNP13`), no matter which key you supply in the bib file. So I have just used the GAPDoc-generated keys for each entry, so that everything is consistent.

I have currently cited all of the new references in a temporary section in the manual, so that all entries are compiled into the bibliography at the end of the manual. This is perhaps controversial.


doc/intro.xml
```diff
@@ -18,6 +18,11 @@ in this package is useful at least for permutation groups, matrix
 groups and projective groups.
 The setup is described in <Cite Key="NS06"/>. <P/>
 
+<Cite Key="LNPS06"/>
+<Cite Key="BLS97"/>
+<Cite Key="JLNP13"/>
+<Cite Key="KS09"/>
```
wilfwilson: These are currently just here so that they appear in the bibliography in the compiled doc.

fingolfin: Makes sense

wilfwilson:
I'm still looking to see if there is any low-hanging 'fruit' (references) in the package for me to add to the bibliography. I think there's just one or two more (currently adding Max's habilitation thesis).

Then I have to decide whether to find the appropriate places in the manual to cite each of these items in this PR. I'm thinking perhaps that should be a separate PR, and for now I can just cite everything in a separate place, or not at all. That way the items can be in the bib file, for people to use if they're also writing documentation (as e.g. @ssiccha is).

wilfwilson:
@fingolfin I've done what I'll do for now. If you're unhappy with the temporary literature section, I can simply remove it. Otherwise I'd be happy for this to be merged.

PR #110, opened 2019-07-04
Improve documentation of SnkSetswrSr method

### LINKED ISSUE #75: Document `jellyfish` methods
Code in `gap/snksetswrsr.gi` mentions jellyfishes. Which paper is this based on?

A reference should be added to the source code.

A link to the paper could be added like in PR #73.

ssiccha:
- Rewrites the documentation for SnkSetswrSr. States more precisely
  what is meant by "large-base primitive group", that is
  which groups are recognised.

Fixes #75.

### DIFF
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -412,7 +412,7 @@ AddMethod(FindHomDbPerm, FindHomMethodsPerm.Imprimitive,
           "for a imprimitive permutation group, restricts to block system");
 AddMethod(FindHomDbPerm, FindHomMethodsPerm.SnkSetswrSr,
           60, "SnkSetswrSr",
-          "tries to find jellyfish" );
+          "recognises large-base primitive permutation groups" );
 AddMethod(FindHomDbPerm, FindHomMethodsPerm.StabilizerChainPerm,
           55, "StabilizerChainPerm",
           Concatenation(
--- a/gap/snksetswrsr.gi
+++ b/gap/snksetswrsr.gi
@@ -474,12 +474,21 @@ RECOG.JellyHomFunc := function(data,el)
 end;
 
 #! @BeginChunk SnkSetswrSr
-#! This method tries to determine whether the input group <A>G</A> is acting
-#! primitively on <M>N</M> points, and
-#! is isomorphic to a large subgroup of <M>H \wr S_r</M> where <M>H</M> is
-#! <M>S_n</M> acting on <M>k</M>-sets and <M>N = {{n \choose k}}^r</M>
-#! and <M>kr > 1</M>. <Q>Large</Q> means that <A>G</A> contains a subgroup
-#! isomorphic to <M>A_n^r</M>. If <A>G</A> is imprimitive then the output is
+#! This method tries to determine whether the input group <A>G</A> is a
+#! large-base primitive group that neither is a symmetric nor an alternating
+#! group in its natural action.
+#! This method is an implementation of <Cite Key="LNPS06"/>.
+#!
+#! A primitive group <M>H</M> acting on <M>N</M> points is called <E>large</E> if
+#! there exist <M>n</M>, <M>k</M>, and <M>r</M> with
+#! <M>N = \{n \choose k\}^r</M>, and up to a permutational isomorphism <M>H</M>
+#! is a subgroup of the product action wreath product <M>S_n \wr S_r</M>,
+#! and an overgroup of <M>(A_n) ^ r</M> where <M>S_n</M> and <M>A_n</M> act on
+#! the <M>k</M>-subsets of <M>\{1, ..., n\}</M>.
+#! This algorithm recognises large primitive groups with <M>r \cdot k > 1</M>
+#! and <M>2 \cdot r \cdot k^2 \le n</M>.
+#!
+#! If <A>G</A> is imprimitive then the output is
 #! <C>NeverApplicable</C>. If <A>G</A> is primitive then
 #! the output is either a homomorphism into the
 #! natural imprimitive action of <A>G</A> on <M>nr</M> points with


ssiccha:
This needs #106 to be merged in first.

gap/snksetswrsr.gi
```diff
@@ -29,7 +29,7 @@ SetInfoLevel( InfoJellyfish, 1 );
 ##
 ##  Computes all [n,k,r] such that N = Binomial(n,k)^r. Returns those
 ##  satisfying k*r > 1 and 2*r*k^2 < n (sufficient condition from
-##  BLS (Lemma 2.10) that ensures that there is a unique minimal and a
+##  BLS97 (Lemma 2.10) that ensures that there is a unique minimal and a
```
wilfwilson: Can you think of a place to (perhaps additionally) cite this paper in the auto doc, so that it appears in the manual?

ssiccha: This looks very much like an implementation detail. I think there is no sensible place in the manual to reference it. :(

gap/perm.gi
```diff
@@ -412,7 +412,7 @@ AddMethod(FindHomDbPerm, FindHomMethodsPerm.Imprimitive,
           "for a imprimitive permutation group, restricts to block system");
 AddMethod(FindHomDbPerm, FindHomMethodsPerm.SnkSetswrSr,
           60, "SnkSetswrSr",
```
fingolfin: So, what does that stand for? Based on what people told me here, the guess would be `\Sym{n over k) \wreath \Sym(r)`.  But that name still is unpronounceable, and that's bad as it inhibits communication on this code. I also don't see why the `Set` needs to be in there -- surely `SnkwrSr` says it, too?

But actually, how about we just rename the method to "LargeBasePerm" or even just "Jeyllyfish"? Both have the great advantage that they are pronounceable. (Of course it's not a requirement for this PR! I just thought you might have some thoughts on this, as you have looked at the code recently).

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,24 @@ RECOG.JellyHomFunc := function(data,el)
...
-#! is isomorphic to a large subgroup of <M>H \wr S_r</M> where <M>H</M> is
-#! <M>S_n</M> acting on <M>k</M>-sets and <M>N = {{n \choose k}}^r</M>
-#! and <M>kr > 1</M>. <Q>Large</Q> means that <A>G</A> contains a subgroup
-#! isomorphic to <M>A_n^r</M>. If <A>G</A> is imprimitive then the output is
+#! This method tries to determine whether the input group <A>G</A> is a
+#! large-base primitive group that neither is a symmetric nor an alternating
+#! group in its natural action.
+#! This method is an implementation of <Cite Key="LNPS06"/>.
+#!
+#! A primitive group <M>H</M> acting on <M>N</M> points is called large if
```
fingolfin: Maybe emphasize the word being defined, like so?
```suggestion
#! A primitive group <M>H</M> acting on <M>N</M> points is called <E>large</E> if
```

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,24 @@ RECOG.JellyHomFunc := function(data,el)
...
+#! there exist <M>n</M>, <M>k</M>, and <M>r</M> with
+#! <M>N = {{n \choose k}}^r</M>
+#! and up to a permutational isomorphism
+#! <M>H</M> is a subgroup of the product action wreath product
+#! <M>S_n \wr S_r</M>
```
fingolfin: The line wrapping is really strange in this section. Usually I try to have lines < 78 columns wide, and then wrapped along those lines... It's OK to deviate, but here it really looks to me as if the text could be reformatted a bit to have slightly more evenly distributed line lengths...

Also, please add a comma here:
```suggestion
#! <M>S_n \wr S_r</M>,
```

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,24 @@ RECOG.JellyHomFunc := function(data,el)
...
+#! and an overgroup of <M>A_n ^ r</M>
+#! where <M>S_n</M> and <M>A_n</M> act on the <M>k</M>-subsets of
+#! <M>{{1, .., n}}</M>.
```
fingolfin:
```suggestion
#! <M>{{1, ..., n}}</M>.
```
Why two braces `{{` and not just one `{` ? (Perhaps some GAPDoc speciality I am not aware of?)

wilfwilson:
Note that #106 is merged, so this can be merged whenever it is ready.

gap/perm.gi
ssiccha: I actually just now thought about how to rename it. My suggestion is `LargeBasePrimitivePerm`.

I'll do the renaming in a different, later PR because I don't want to mess with other people's PRs.

fingolfin:
Needs to be rebased.

gap/snksetswrsr.gi
ssiccha: One brace `{` is not printed by GAPDoc on my machine. I guess `{{` is the escaped version of `{`.

gap/perm.gi
fingolfin: I am pretty sure that's not the way you should do it; those formulas are meant to adhere to LaTeX conventions, so you should be using `\{` and `\}` -- see also https://www.gap-system.org/Manuals/pkg/GAPDoc-1.6.1/doc/chap3.html#X7ABF42328467E966

fingolfin: OK

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,21 @@ RECOG.JellyHomFunc := function(data,el)
...
-#! and <M>kr > 1</M>. <Q>Large</Q> means that <A>G</A> contains a subgroup
-#! isomorphic to <M>A_n^r</M>. If <A>G</A> is imprimitive then the output is
+#! This method tries to determine whether the input group <A>G</A> is a
+#! large-base primitive group that neither is a symmetric nor an alternating
+#! group in its natural action.
+#! This method is an implementation of <Cite Key="LNPS06"/>.
+#!
+#! A primitive group <M>H</M> acting on <M>N</M> points is called <E>large</E> if
+#! there exist <M>n</M>, <M>k</M>, and <M>r</M> with
+#! <M>N = {{n \choose k}}^r</M>, and up to a permutational isomorphism <M>H</M>
```
fingolfin: I am pretty sure that's not the way you should do it; those formulas are meant to adhere to LaTeX conventions, so you should be using `\{` and `\}` -- see also https://www.gap-system.org/Manuals/pkg/GAPDoc-1.6.1/doc/chap3.html#X7ABF42328467E966

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,21 @@ RECOG.JellyHomFunc := function(data,el)
...
+#! is a subgroup of the product action wreath product <M>S_n \wr S_r</M>,
+#! and an overgroup of <M>(A_n) ^ r</M> where <M>S_n</M> and <M>A_n</M> act on
+#! the <M>k</M>-subsets of <M>{1, ..., n}}</M>.
```
fingolfin: Braces again

gap/perm.gi
fingolfin: Although I don't see which PR you are afraid of conflicting with?

gap/snksetswrsr.gi
```diff
@@ -474,12 +474,21 @@ RECOG.JellyHomFunc := function(data,el)
...
+#! This algorithm recognises large primitive groups with <M>r*k > 1</M>
```
fingolfin: Perhaps use `\cdot` here and in the next line?

gap/perm.gi
ssiccha: Ah, I wanted to do the renaming as suggested [in this hack.md](https://hackmd.io/TJflWtgcSFmSd2WHwneMtA) all at once. That's what I meant with "in a later PR".

gap/snksetswrsr.gi
ssiccha: Ok. I'll also exchange the `*` for `\cdot`.

gap/snksetswrsr.gi
ssiccha: Yepp

ssiccha:
Rebased and updated.

gap/perm.gi
fingolfin: I really don't see a value in "doing it all at once". To the contrary: It means that changes which might be uncontroversial would be delayed by other changes that *are* controversial.

Renaming this method is uncontroversial, clearly fits into this PR, and is trivial to exclude from "that HackMD". So you could just have done it here... As it is, I'll do it myself then.

ssiccha:
I guess you're right. I was thinking: "If I rename it now I could as well do it in a script. If I do it in a script I might as well do it for everything at once."

PR #112, opened 2019-07-04
Refactor CalcNiceGensHomNode

ssiccha:


wilfwilson:
If you want to refactor this, I would suggest the following:
```gap
InstallGlobalFunction( CalcNiceGensHomNode,
  # function for the situation on a homomorphism node (non-Leaf):
  function(ri, origgens)
    local preimagesNiceGensOfFactor, kernelGens;
    preimagesNiceGensOfFactor := CalcNiceGens(RIFac(ri), origgens);
    # Is there a non-trivial kernel?
    if HasRIKer(ri) and RIKer(ri) <> fail then
        kernelGens := ResultOfStraightLineProgram(gensNslp(ri), origgens);
        Append(preimagesNiceGensOfFactor, CalcNiceGens(RIKer(ri), kernelGens));
    fi;
    return preimagesNiceGensOfFactor;
  end );
```
Removes variables that are used only once, and avoids having to do anything extra when there's a trivial kernel.

gap/base/recognition.gi
```diff
@@ -678,18 +678,19 @@ InstallGlobalFunction( CalcNiceGensGeneric,
 InstallGlobalFunction( CalcNiceGensHomNode,
   # function for the situation on a homomorphism node (non-Leaf):
   function(ri,origgens)
-    local origkergens,rifac,riker,pregensfactor;
+    local gensOfKernel,rifac,riker,preimagesOfNiceGensOfFactor,
+    niceGensOfKernel;
```
```suggestion
          niceGensOfKernel;
```

gap/base/recognition.gi
```diff
@@ -678,18 +678,19 @@ InstallGlobalFunction( CalcNiceGensGeneric,
...
     # Is there a non-trivial kernel?
     rifac := RIFac(ri);
+    preimagesOfNiceGensOfFactor := CalcNiceGens(rifac,origgens);
     if HasRIKer(ri) and RIKer(ri) <> fail then
-        pregensfactor := CalcNiceGens(rifac,origgens);
         riker := RIKer(ri);
-        origkergens := ResultOfStraightLineProgram( gensNslp(ri), origgens );
-        return Concatenation( pregensfactor,
-                              CalcNiceGens(riker,origkergens) );
+        gensOfKernel := ResultOfStraightLineProgram( gensNslp(ri), origgens );
```
fingolfin: Looking at the def for `gensNslp`,  I saw this:
```
        SetgensNslp(ri,SLPOfElms(gensN(ri)));
```
Which made me wonder why one can't just use something like `List( gensN(ri), StripMemory )` here..

The answer being (I think) that the output here really needs to depend on `origgens`, and these might not be original gens we worked, they could be arbitrary stand-ins which satisfy the same relations. Perhaps this could be made clearer somehow by some comments?

wilfwilson:
@fingolfin In the case that you look at this again, I made a suggestion for a different possible refactoring. But I did it in a comment in rather than a review, so you possibly missed it.

fingolfin:
@wilfwilson ah, thanks! Yeah, OK, I like that, but would do it yet again differently: because after all, we are computing the nice gens of the current node here, so why not call them that? I.e.:
```gap
InstallGlobalFunction( CalcNiceGensHomNode,
  # function for the situation on a homomorphism node (non-Leaf):
  function(ri, origgens)
    local nicegens, kernelGens;
    # compute preimages of the nicegens of the factor group
    nicegens := CalcNiceGens(RIFac(ri), origgens);
    # Is there a non-trivial kernel? then add its nicegens
    if HasRIKer(ri) and RIKer(ri) <> fail then
        # we cannot just use gensN(RIKer(ri)) here, as those values are defined
        # relative to the original generators we used during recognition; but the
        # origgens passed to this function might differ, e.g. they might be conjugates
        # of the original generators (they just have to satisfy the same relations as
        # those, not be identical to them).
        kernelGens := ResultOfStraightLineProgram(gensNslp(ri), origgens);
        Append(nicegens, CalcNiceGens(RIKer(ri), kernelGens));
    fi;
    return nicegens;
  end );
```

wilfwilson:
I like that @fingolfin

ssiccha:
> I like that @fingolfin

Me too. I'll update the PR accordingly.

ssiccha:
```
        # we cannot just use gensN(RIKer(ri)) here, as those values are defined
        # relative to the original generators we used during recognition; but the
        # origgens passed to this function might differ, e.g. they might be conjugates
        # of the original generators (they just have to satisfy the same relations as
        # those, not be identical to them).
```
Do you know a situation where this could happen @fingolfin? I  wasn't easily able to come up with one and I think it would be nice if that could also be added to the comment.

ssiccha:
Rebased and updated.

fingolfin:
@ssiccha we looked at examples for that a few days ago, didn't we? See `RECOG.CalcNiceGensKnownNilpotent`.

ssiccha:
@fingolfin I had a look again at `RECOG.CalcNiceGensKnownNilpotent`. It looks like the elements of `b` are conjugates of the `origgens` but actually the `^` is used here as powering by an integer, not as conjugation.

I suggest to shorten the comment to this:
```
        # we cannot just use gensN(RIKer(ri)) here, as those values are defined
        # relative to the original generators we used during recognition; but the
        # origgens passed to this function might differ
```

ssiccha:
I just added a new commit to this PR.

Note: squash the commits when merging them in.

ssiccha:
Rebased onto master.

PR #119, opened 2019-07-04
Add some broken tests which test many subgroups

ChrisJefferson:
This adds some broken tests, mainly so there is examples of how to use some of the advanced testing functions which can be merged later, when they work.


wilfwilson:
@ChrisJefferson Not a problem, but in case you didn't realise: you pushed your branch to the `gap-packages` repo rather than your own fork.

PR #120, opened 2019-07-04
Some tweaks to the early parts of the manual

wilfwilson:


doc/methsel.xml
```diff
@@ -5,7 +5,7 @@
 -->
 
 <Chapter Label="methsel">
-<Heading>Method Selection</Heading>
+<Heading>Method selection</Heading>
```
DominikBernhardt: Is there a reason why above you refer to this as `Method selection system` but left the title of method selection here?

DominikBernhardt: I think we should make clear that this is different to the GAP method selection. Was that your intention?

wilfwilson: My intention was to make the capitalisation of the chapter titles the same. I think it’s fine to refer to the ‘method selection system’ while sometimes using the name ‘method selection’, especially as a title. I am sure that the documentation clearly communicates that it is a custom system; if not that is certainly something I will address when I look at that chapter.

PR #128, opened 2019-07-05
WIP: Improve function names

ssiccha:
This is very much work in progress and in an early stage.

The new names we suggest can be found [on this hack.md](https://hackmd.io/TJflWtgcSFmSd2WHwneMtA).

During our work on this we thoroughly read chapter 3 of the manual and got some ideas on how to improve it. That's going to be in future PRs.


doc/renaming.xml
```diff
@@ -0,0 +1,7 @@
+<Chapter Label="Renaming">
+<Heading>Renaming</Heading>
+
+In this chapter a table of renamed methods is included.
+The old names are explained in <Cite Key="IssacRecog"/>.
+
+<#Include SYSTEM "renaming_table.xml">
```
ssiccha: We call auto generated files `_NAME.xml` that is they start with an underscore. Files inside `doc/` of this form are ignored by git per our `.gitignore` file.

doc/renaming_table.xml
```diff
@@ -0,0 +1,30 @@
+<Table Align="|l|l|">
```
ssiccha: This file is an automatically generated file so it must not be added to the commit. You can do `git rm doc/renaming_table.xml` to remove it.

rename.g
```diff
@@ -0,0 +1,324 @@
...
+## in the documentation as 'X_1 -> X_n'.
+## --------------------------------------------------------------------------
+## EXAMPLE:
+## 
+## renamings_to_execute.csv :
+## OLD NAME,NEW NAME,TYPE
+## undesciptive,descriptive,Attr
+## 
+## First add and commit the renamings_to_execute.csv file. -- Commit 1
+## Then discuss on github if the suggested names are suitable.
```
ssiccha: In principle people can also directly create what you call `Commit 1` and `Commit 2` and push that to github. So I'd remove this line and change the next line to
```## Then run `gap rename.g` from within the package directory.```

File: rename.g [hunk 4]
```diff
@@ -0,0 +1,324 @@
...
+## After approval run rename.g from within the package directory.
+## Run tst/testquick.g or tst/testall.g and ensure everything is still working.
+## On failure you can always roll back to Commit 1, i.e. the version before 
+## applying the renamings. 
+## On success add and commit the changed files. -- Commit 2
+##
+#############################################################################
+
+DeclareSynonym("Yes", true);
```
ssiccha: This is not necessary and makes `Yes` read-only. Thus it might in principle clash with e.g. somebody's custom `.gaprc` file.

FriedrichRober:
In reply to @ssiccha:
The renamings_history.csv is in fact only important for the second application of rename.g, i.e.
for people who use the recog package as a dependency. They need to be able to replicate all renamings from any starting point of the renamings_history.csv.
For example if we have a renaming X -> Y -> Z, there may be users who have already renamed function X to Y, but not Y to Z. If we only store X -> Z, there is no possibility for them to use rename.g to automatically apply the renaming from Y -> Z.

rename.g
FriedrichRober: You are right. I suggest to use true/false instead of Yes/No as this should be clear enough for users of this file.

rename.g
```diff
@@ -0,0 +1,324 @@
...
+## The further explanations apply if the file is not empty.
+## 
+## First this file checks if the renamings_to_execute.csv file is correct.
+## The renamings_to_execute.csv file has the following form:
+##     - First column 'OLD NAME' contains the old name.
+##     - Second column 'NEW NAME' contains the new name that is used now.
+##     - Third column 'TYPE' contains the type used for Ref in GAPDoc.
+## Then further checks are done and on failure Errors/Warnings are printed.
+## Primarily they are done to ensure that the renamings_history.csv file 
+## always has duplicate-free columns after merging.
```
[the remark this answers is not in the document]
FriedrichRober: With merging I mean the generation of the `renamings_history.csv` after concatenation with the `renamings_to_execute.csv`.
With duplicate-free I wanted to say that there cannot appear an entry `X `twice in a `Old/New Name` column. Maybe this is a little bit too restrictive and we should allow that. Then I would need to change the implementation of the simplification of the csv, where we reduce `X_1 -> ... -> X`_n to `X_1 -> X_n`. Nevertheless I would say that duplicates should still throw a Warning instead of an Error in most cases. We can discuss that for each case individually, if you wish.

ssiccha:
@FriedrichRober I see. That's very nice. :) 

In general, since we don't know whether there actually is somebody who needs that use case you could have left that out to implement it only if needed. I could then have created the history file manually from the commit history. It's better to first stick with the things that we know we'll need since otherwise things can get out of hand quickly.

rename.g
```diff
@@ -0,0 +1,324 @@
...
+
+        Print(message,"\n\tThe word \'",word,"\' is located \n\tin line ",posA+1," in column \'",colNameA,"\' in file \'",fileA,"\' and\n\tin line ",posB+1," in column \'",colNameB,"\' in file \'",fileB,"\'\n");
+    od;
+    return failedCheck;
+end;
+
+# Returns true on failed check. Helper method of CheckRenamingTable.
+# Check fails if grep returns failureInt on search of a word in col.
+# failureInt is either 0 or 1.
+CheckExistenceOfWord := function(col, colName, failureInt, message)
```
ssiccha: You can let the function return whether  a `word` was found and remove the `failureInt` variable. That makes the function easier to understand.

FriedrichRober: Yes, but then I would need to store somewhere the words which are (found/not found) for the error message, which is printed at the moment from within the function depending on the `failureInt`. I think this would be even less readable.

ssiccha: Ah, true. You could in principle return the list of found words and let the rest be handled by the calling function. But let's not get too concerned with details. Leave the function like it is if you want to.

rename.g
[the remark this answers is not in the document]
ssiccha: > Nevertheless I would say that duplicates should still throw a Warning instead of an Error in most cases.

Jepp. I actually think a warning is preferable in all cases, at least in the "duplicates" cases.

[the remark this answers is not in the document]
ssiccha: > Then I would need to change the implementation of the simplification of the csv, where we reduce X_1 -> ... -> X_n to X_1 -> X_n.

That's about "use case 2" though, right? Then I'd suggest let's ignore the possibility of duplicates there (in the simplification function) for now. We can still take care of duplicates _if_ they really appear. After all if a PR introduces a duplicate which doesn't make any sense, that PR shouldn't make it through the reviewing process.

renamings_history.csv
```diff
@@ -0,0 +1 @@
+OLD_NAME,NEW NAME,TYPE
```
ssiccha:
```suggestion
OLD NAME,NEW NAME,TYPE
```

renamings_to_execute.csv
```diff
@@ -0,0 +1 @@
+OLD NAME,NEW NAME,TYPE
```
ssiccha: Delete this file from the commit.

RENAME.md
```diff
@@ -0,0 +1,82 @@
...
+Now let us discuss the case where the `renamings_to_execute.csv` file is empty.
+Then the `rename.g` file attempts to do the renamings specified
+in the `renamings_history.csv` file.
+The replacements are executed as described in the first case.
+
+## EXAMPLE
+
+This example covers the use case for developers of the recog package.
+
+Change the `renamings_to_execute.csv` file, so it looks like this:
```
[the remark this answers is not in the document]
ssiccha: I've thought about the `renamings_to_execute.csv`. I don't think it's a good idea to have it lying around in the root directory of the package. It may be a little less comfortable but it's cleaner this way.

RENAME.md
```diff
@@ -0,0 +1,82 @@
...
+
+    OLD NAME,NEW NAME,TYPE
+    FindHomDbPerm,MoreDescriptiveNewName,Attr
+
+Inside the package directory do:
+
+    git add renamings_to_execute.csv
+    git commit -m "Rename Functions"
```
ssiccha: Remove these two lines.

RENAME.md
```diff
@@ -0,0 +1,82 @@
...
+
+    gap tst/testall.g
+
+On failure you can always roll back to the last commit, i.e. the version before
+applying the renamings. For example one could type in the command `git reset --hard`.
+
+On success add the changed files:
+
+    git add *
+    git commit --amend
```
ssiccha:
```suggestion
    git reset renamings_to_execute.csv
    git commit
```

doc/renaming_table.xml
```diff
@@ -0,0 +1,10 @@
...
+<Caption> renaming of functions</Caption>
+<HorLine/>
+<Row>
+	<Item> <E>OLD NAME</E> </Item>
```
ssiccha:
```suggestion
	<Item> <E>Old Name</E> </Item>
```
Not using all caps looks nicer IMO.

doc/renaming_table.xml
```diff
@@ -0,0 +1,10 @@
...
+	<Item> <E>NEW NAME</E> </Item>
```
ssiccha:
```suggestion
	<Item> <E>New Name</E> </Item>
```

rename.g
```diff
@@ -0,0 +1,259 @@
...
+        return false;
+    elif input = true then
+        return true;
+    else
+        Print("Input is not valid.\n");
+        return GetInputFromUser(message);
+    fi;
+end;
+
+# Returns true on failed check. Helper method of CheckRenamingTable.
```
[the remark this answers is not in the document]
ssiccha: Maybe you can even get rid of `isFirstInHistory`, see below.

RENAME.md
```diff
@@ -0,0 +1,82 @@
...
+Then further sanity checks are done and on failure Errors/Warnings are printed.
+After successful checks all matches on word boundaries of the old name get
+replaced with the new name. The replacements are executed in all files with
+extensions .g, .gd, .gi, .tst, and .xml found in the directory tree rooted at '.'
+whereas all files in the .git folder are ignored.
+Afterwards the rows from the `renamings_to_execute.csv` file are added to
+the `renamings_history.csv` file. Then the renaming table for the
+documentation is generated from the `renamings_history.csv` file whereby
+renamings of the form `X_1 -> X_2 -> ... -> X_n` are printed
+in the documentation as `X_1 -> X_n`.
```
ssiccha:
```suggestion
in the documentation as `X_1 -> X_n`. Then the documentation of the recog package is regenerated.
```

renamings_history.csv
FriedrichRober: I have not noticed this before, but this seems to be a bug in gap. When the csv file is auto generated, `OLD_NAME` is written instead of `OLD NAME`. I will look into this.
To test this, type in the following commands in gap:

```
R := [rec(OLD_NAME := "old", NEW_NAME := "new", TYPE := "func")];
PrintCSV("renamings_history.csv", R, ["OLD_NAME", "NEW_NAME", "TYPE"]);
```

ssiccha: Yup, looks like a bug. Printing `OLD_NAME` is correct though. Instead its printing `NEW NAME` when it should be printing `NEW_NAME`. Nice catch. :)

Did you already report this?

ssiccha:
Closed in favor of #171.

PR #131, opened 2019-07-08
Improve documentation of functions in code and manual II

PaulaHaehndel:
Most functions in gap/perm.gi are commented in the code now. In case they appear in the documentation they also have now an AutoDoc entry.
Additionally some functions in gap/matrix.gi got a documenting comment.

Done by @monschau, @Munsee and me (the code part of the documentation group) on the SummerSchoolMGR.

### DIFF
--- a/gap/matrix.gi
+++ b/gap/matrix.gi
@@ -626,6 +626,7 @@ end;
 #end;
 #
 RECOG.ExtractLowStuff := function(m,layer,blocks,lens,canbas)
+  #canbas ... canonical basis
   local block,i,j,k,l,pos,v,what,where;
   v := ZeroVector(lens[layer],m[1]);
   pos := 0;
@@ -642,6 +643,9 @@ RECOG.ExtractLowStuff := function(m,layer,blocks,lens,canbas)
       od;
   od;
   if canbas <> fail then
+      # needed because we assume for example in
+	  # SLPforElementFuncsMatrix.LowerLeftPGroup that we work
+	  # over a field of order p (not a p power)
       return BlownUpVector(canbas,v);
   else
       return v;
@@ -763,6 +767,11 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
     return true;
   end );
 
+# computes a straight line program (SLP) of <g> for a p-group in <ri> (that
+# correspond to a lower left triangular matrix;
+# The SLP is constructed by using the elementar matrices to cancel out the
+# entries in <g>. The coefficents of the process create the SLP.
+# TODO Max H. wants to rewrite the code
 SLPforElementFuncsMatrix.LowerLeftPGroup := function(ri,g)
   # First project and calculate the vector:
   local done,h,i,l,layer,pow;
@@ -798,7 +807,7 @@ end;
 #! @BeginChunk LowerLeftPGroup
 #! This method is only called by a hint from <C>BlockLowerTriangular</C>
 #! as the kernel of the homomorphism mapping to the diagonal blocks.
-#! The method uses the fact the this kernel is a <M>p</M>-group where
+#! The method uses the fact that this kernel is a <M>p</M>-group where
 #! <M>p</M> is the characteristic of the underlying field. It exploits
 #! this fact and uses this special structure to find nice generators
 #! and a method to express group elements in terms of these.
@@ -846,7 +855,12 @@ FindHomMethodsMatrix.GoProjective := function(ri,G)
 end;
 
 #! @BeginChunk KnownStabilizerChain
-#! TODO. use an already known stabilizer chain for this group
+#! If a stabilizer chain is already known, then the kernel node is given
+#! knowledge about this known stabilizer chain, and the factor node is told to
+#! use homomorphism methods from the database for permutation groups.
+#! If a stabilizer chain of a parent node is already known this is used for
+#! the computation of a stabilizer chain of this node. This stabilizer chain
+#! is then used in the same way as above.
 #! @EndChunk
 FindHomMethodsMatrix.KnownStabilizerChain := function(ri,G)
   local S,hom;
--- a/gap/matrix.gi
+++ b/gap/matrix.gi
@@ -626,7 +626,6 @@ end;
 #end;
 #
 RECOG.ExtractLowStuff := function(m,layer,blocks,lens,canbas)
-  #canbas ... canonical basis
   local block,i,j,k,l,pos,v,what,where;
   v := ZeroVector(lens[layer],m[1]);
   pos := 0;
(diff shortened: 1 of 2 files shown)


gap/matrix.gi
```diff
@@ -763,6 +767,11 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
     return true;
   end );
 
+# computes a straight line program (SLP) of <g> for a p-group in <ri> (that
+# correspond to a lower left triangular matrix;
```
fingolfin:
```suggestion
# corresponds to a lower triangular matrix group;
```
or: "... that consists of lower triangular matrices", or  "... a matrix group consisting of lower triangular matrices" or so

gap/matrix.gi
```diff
@@ -763,6 +767,11 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
     return true;
   end );
 
+# computes a straight line program (SLP) of <g> for a p-group in <ri> (that
```
fingolfin:
```suggestion
# computes a straight line program (SLP) for an element <g> of a p-group in <ri> that
```

gap/perm.gi
```diff
@@ -130,7 +134,13 @@ FindHomMethodsPerm.PcgsForBlocks := function(ri,G)
 end;
 
 #! @BeginChunk BalTreeForBlocks
-#! TODO
+#! This method creates a balanced composition tree for the kernel of an
+#! imprimitive group. This is guaranteed as
+#! <C>FindHomMethodsPerm.BalTreeForBlocks</C> is
+#! just called from <C>FindHomMethodsPerm.Imprimitive</C> and itself.
+#! The homomorphism for the split in the composition tree used is
+#! induced by the action of G on
```
fingolfin: Not very important, but: elsewhere you wrote `<A>G</A>`, though I guess one could argue that this is not correct (it is not an argument); another alternative would be `<M>G</M>`. I guess we just should settle on one.

gap/perm.gi
```diff
@@ -130,7 +134,13 @@ FindHomMethodsPerm.PcgsForBlocks := function(ri,G)
 end;
 
 #! @BeginChunk BalTreeForBlocks
-#! TODO
+#! This method creates a balanced composition tree for the kernel of an
+#! imprimitive group. This is guaranteed as
+#! <C>FindHomMethodsPerm.BalTreeForBlocks</C> is
+#! just called from <C>FindHomMethodsPerm.Imprimitive</C> and itself.
```
fingolfin: I would recommend to use proper refs here, something like `<Ref Subsect="Imprimitive"/>` or so should do it.

gap/matrix.gi
```diff
@@ -642,6 +643,9 @@ RECOG.ExtractLowStuff := function(m,layer,blocks,lens,canbas)
       od;
   od;
   if canbas <> fail then
+      # needed because we assume for example in
+	  # SLPforElementFuncsMatrix.LowerLeftPGroup that we work
+	  # over a field of order p (not a p power)
```
fingolfin: Something is wrong with the indent here -- tabs vs spaces?

gap/matrix.gi
```diff
@@ -626,6 +626,7 @@ end;
 #end;
 #
 RECOG.ExtractLowStuff := function(m,layer,blocks,lens,canbas)
+  #canbas ... canonical basis
```
fingolfin: canonical basis of what, though?

gap/matrix.gi
```diff
@@ -763,6 +767,12 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
     return true;
   end );
 
+# computes a straight line program (SLP) for an element <g> of for a p-group
```
fingolfin:
```suggestion
# computes a straight line program (SLP) for an element <g> of a p-group
```

gap/matrix.gi
```diff
@@ -763,6 +767,12 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
...
+# in <ri> (that corresponds to a matrix group consisting of lower triangular
```
fingolfin: Here a parenthesis opens that never is closed.

Also, is the p-group really "in <ri>"? Perhaps the p-group corresponds to <ri>, or is described by <ri>, or... ?

gap/matrix.gi
```diff
@@ -763,6 +767,12 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
...
+# matrices.
+# The SLP is constructed by using the elementar matrices to cancel out the
```
fingolfin:
```suggestion
# The SLP is constructed by using the elementary matrices to cancel out the
```
Though actually that's not correct, the matrices used for reduction are in general not elementary. Perhaps like this:
```suggestion
# The SLP is constructed by using matrices forming a pcgs (polycyclic generating sequence)
# of the group to clean out the entries of the input matrix in the lower left part entry by entry.
```

gap/matrix.gi
```diff
@@ -763,6 +767,12 @@ InstallGlobalFunction( FindKernelLowerLeftPGroup,
...
+# entries in <g>. The coefficents of the process create the SLP.
+# TODO Max H. wants to rewrite the code
```
fingolfin:
```suggestion
# TODO Max H. wants to rewrite the code to work row-by-row
# instead of blockwise; that should result in both simple and faster code.
```

gap/perm.gi
```diff
@@ -130,7 +134,12 @@ FindHomMethodsPerm.PcgsForBlocks := function(ri,G)
 end;
 
 #! @BeginChunk BalTreeForBlocks
-#! TODO
+#! This method creates a balanced composition tree for the kernel of an
+#! imprimitive group. This is guaranteed as the method is just called
+#! from <C>Imprimitive</C> (see <Ref Subsect="Imprimitive"/>) and itself.
```
fingolfin:
```suggestion
#! from <Ref Subsect="Imprimitive" Style="Text"/> and itself.
```

gap/perm.gi
```diff
@@ -113,7 +113,11 @@ FindHomMethodsPerm.Imprimitive :=
   end;
 
 #! @BeginChunk PcgsForBlocks
-#! TODO
+#! This method is called after a hint is set in
+#! <C>FindHomMethodsPerm.Imprimitive</C>. Therefore the group <A>G</A> preserves
```
fingolfin: I am torn here... I want to say that you could use `<Ref Subsect="Imprimitive" Style="Text"/>` here instead -- but then for somebody reading the code, it would be beneficial to see `FindHomMethodsPerm.Imprimitive`. Hrm. Perhaps we can find a way to allow using the label `FindHomMethodsPerm.Imprimitive`, then we could have the advantages of both.

For now, it is perfectly fine to leave this as-is.

gap/matrix.gi
PaulaHaehndel: I don't know.

PaulaHaehndel: I think I understood less then I assumed, about the variable.

gap/perm.gi
PaulaHaehndel: I changed the `BalTreeForBlocks` acording to the fact that you should also be able to read and understand it in the code. So it is now in kind of the same state as this bit.

PR #134, opened 2019-07-08
Documentation: add more information about feedback options

wilfwilson:
Mostly this is about advertising the mailing list.

I've also added something about the required packages to the installation page of the documentation, which was taken from the README. In doing so, I decided to have the required packages listed in a consistent (alphabetical) order.

### DIFF
--- a/doc/install.xml
+++ b/doc/install.xml
@@ -8,14 +8,24 @@ This is the chapter of the documentation describing the installation.
 <Heading>Installation of the <Package>recog</Package> package</Heading>
 <Index><Package>recog</Package></Index>
 
-To install this  package, just extract the package's  archive file to
-the  GAP  <F>pkg</F>  directory.<P/>
+To install this package, just extract the package's archive file to the &GAP;
+<F>pkg</F> directory.<P/>
 
-By default, the <Package>recog</Package> package is not automatically
-loaded by &GAP; when it is installed. You must load the package with
+If the <Package>recog</Package> package is not automatically loaded
+when &GAP; is started, then you must load the package with
 <C>LoadPackage("recog");</C> before its functions become available.<P/>
 
-<!-- TODO Write something concerning the required packages -->
+Note that the <Package>recog</Package> package needs the
+<Package>AtlasRep</Package>,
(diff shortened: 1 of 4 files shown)


doc/install.xml
```diff
@@ -8,14 +8,24 @@ This is the chapter of the documentation describing the installation.
...
 
-To install this  package, just extract the package's  archive file to
-the  GAP  <F>pkg</F>  directory.<P/>
+To install this package, just extract the package's archive file to the &GAP;
+<F>pkg</F> directory.<P/>
 
-By default, the <Package>recog</Package> package is not automatically
-loaded by &GAP; when it is installed. You must load the package with
+By default, the <Package>recog</Package> package is not automatically loaded
+by &GAP; when it is installed. You must load the package with
```
fingolfin: That actually doesn't seem to be accurate anymore, at least for me, it does get loaded because `atlasrep` 2.x suggests it... I am not exactly about that, though, but that's how it is.

doc/intro.xml
```diff
@@ -54,11 +54,17 @@ usage of this package.<P/>
...
 
-If you have any bug reports, feature requests, or suggestions about the
-package, please use our
-<URL Text="issue tracker">https://github.com/gap-packages/recog/issues</URL>.
+If you have any bug reports, feature requests, or suggestions, then please
+tell us via the <URL Text="issue tracker on
+GitHub">https://github.com/gap-packages/recog/issues</URL>. <P/>
+
+In addition, the <Package>recog</Package> package has a mailing list, at
+<URL>recog@gap-system.org</URL>, which can be used for holding discussions,
```
fingolfin: Does GAPDoc handle emails like that? Think is, `recog@gap-system.org` is not a valid URL, but `mailto: recog@gap-system.org` is. So perhaps this needs to be `<URL Text="recog@gap-system.org"> recog@gap-system.org</URL>`?

wilfwilson: Yes it wasn't working.

doc/install.xml
wilfwilson: I've changed it to something a bit more general.

PR #135, opened 2019-07-08
PackageInfo.g: correct last release date

wilfwilson:
Sure, a release date is kind of irrelevant for a development version. But I'd still feel happier with the date being the correct last release date rather than one with the wrong year.

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -8,7 +8,7 @@ SetPackageInfo( rec(
 PackageName := "recog",
 Subtitle := "A collection of group recognition methods",
 Version := "1.3.2dev",
-Date := "15/04/2018", # dd/mm/yyyy format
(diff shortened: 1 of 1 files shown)


PackageInfo.g
```diff
@@ -8,7 +8,7 @@ SetPackageInfo( rec(
 PackageName := "recog",
 Subtitle := "A collection of group recognition methods",
 Version := "1.3.2dev",
-Date := "15/04/2018", # dd/mm/yyyy format
+Date := "15/04/2019", # dd/mm/yyyy format
```
fingolfin: Ouch, so 1.3.2 had a bad release date and I never noticed? :-(. Really need a check in ReleaseTools to check if the release date is in the past...

wilfwilson:
Yep, might as well.

fingolfin:
Thanks!

PR #139, opened 2019-07-09
More homomrangecheck

ChrisJefferson:
This is an extension of #98, which adds many more homomrangecheck calls, and also fixes an SLP bug in An and Sn.

WIth this (as the tests show), we can now hit all Subgroups of Sym(10), and all transitive groups of size 16. These numbers are chosen just to be large enough and run in "reasonable" time.

If we push much larger than this we start failing failures, but those are due to only recognising a subgroup instead of the whole group, not SLP/element problems.

This contains #98

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -528,7 +528,11 @@ InstallGlobalFunction( RecogniseGeneric,
               counter,").");
         fi;
         Add(depth,'F');
-        Assert(2, ForAll(GeneratorsOfGroup(H), x->ValidateHomomInput(ri, x)));
+        if ForAny(GeneratorsOfGroup(H), x->not ValidateHomomInput(ri, x)) then
+            # Our group fails to contain some of the generators of H!
+            return fail;
+        fi;
+
         rifac := RecogniseGeneric(
                   Group(List(GeneratorsOfGroup(H), x->ImageElm(Homom(ri),x))),
                   methodsforfactor(ri), depth, forfactor(ri) ); # TODO: change forfactor to hintsForFactor??)
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -50,6 +50,7 @@ FindHomMethodsPerm.NonTransitive :=
     o := Orb(G,la,OnPoints);
     Enumerate(o);
     hom := OrbActionHomomorphism(G,o);
+    Setvalidatehomominput(ri, {ri,p} -> ForAll(o, x -> (x^p in o)));
     SetHomom(ri,hom);
     return Success;
   end;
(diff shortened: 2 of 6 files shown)


fingolfin:
Nice, thanks! I'll wait for PR #98 to be merged, then this can be rebased and the new changes are easier to review.

ChrisJefferson:
This has been rebased on top of master (and the changed/merged #98)

gap/base/recognition.gi
```diff
@@ -528,7 +528,11 @@ InstallGlobalFunction( RecogniseGeneric,
               counter,").");
         fi;
         Add(depth,'F');
-        Assert(2, ForAll(GeneratorsOfGroup(H), x->ValidateHomomInput(ri, x)));
+        if ForAny(GeneratorsOfGroup(H), x->not ValidateHomomInput(ri, x)) then
+            # Our group fails to contain some of the generators of H!
+            return fail;
+        fi;
```
fingolfin: Hrm, I am not completely sure if I believe that this is the correct thing to do at this point. In any case, I think it's closely related to the mandarin work. I'll have to think about this for a little bit (but not right now).

That said, the worst this can do is break something that worked before; the best is that it prevents some incorrect results; so overall, even if it needs to be done slightly differently in the end, I guess it won't hurt as long as the tests keep passing :-)

gap/perm.gi
```diff
@@ -159,6 +161,10 @@ FindHomMethodsPerm.BalTreeForBlocks := function(ri,G)
   upperhalf := blocks{[cut+1..nrblocks]};
   o := Concatenation(upperhalf);
   hom := ActionHomomorphism(G,o);
+
+  # Make a set so checking validatehomominput is faster
+  seto := Immutable(Set(o));
```
fingolfin: Actually, often computing the `ActionHomomorphism` is also faster if one gives it a set. So perhaps just redefine `o` above?

ChrisJefferson: Actually, I did that and it broke other things, which assumed that `o` was still `Concatenation(upperhalf)`. I could add a comment to say why that wasn't changed.

fingolfin: Ah, I see. Well, such a comment would then indeed be helpful :-)

gap/base/recognition.gi
ChrisJefferson: This certainly should be fixed with mandarins, we shouldn't really get to this place -- it means (I think) we've decided on a homomorphism where the generators of the group aren't valid inputs.

Might be worth thinking about if the set of mandarins should always include the generators. On one hand they are "low quality", on the other hand any recognised group really better include the generators the user gave :)

ChrisJefferson:
tweaked/tided

PR #142, opened 2019-07-09
Simplify and speed up IsPowerOfPrime

wilfwilson:
There's nothing wrong with the old code for `IsPowerOfPrime`, but I expected that GAP would have a built-in function to do the same, and being GAP, it would probably be faster. That seemed to be near enough the case: there exists `IsPrimePowerInt`, which while much more complicated than this function, seems to perform well. Using it, I get the following speedup:
```
gap> Number([1 .. 10000000], n -> IsPowerOfPrimeOLD(n, 2)); time;
23
17340
gap> Number([1 .. 10000000], n -> IsPowerOfPrimeNEW(n, 2)); time;
23
2624
```
(Note that `IsPowerOfPrime` is only ever called for the prime `2` in recog. But the new one is faster for everything I tested.)

In the spirit of keeping `recog` simple, I propose that we merge this PR and reduce the amount of code in the package.

### DIFF
--- a/gap/classical.gi
+++ b/gap/classical.gi
@@ -34,23 +34,8 @@
 BindGlobal( "ClassicalMethDb", [] );
 #
 ###
-## Test whether n is a power of the prime p
-##
-IsPowerOfPrime := function( n, p )
-
-    local x;
-
-    if n <= 0 then return false; fi;
-
-    repeat
-        x := QuotientRemainder( n, p );
-        if x[2] <> 0 then return false; fi;
-        n := x[1];
-    until n = 1;
-
-    return true;
-
-end;
+# Test whether <n> (assumed integer) is a power of <p> (assumed positive prime)
+IsPowerOfPrime := {n, p} -> n > 1 and p ^ PVALUATION_INT(n, p) = n;
 
 
 # Check if m > 5 and the order of a basic lppd(d,q;e) element


DominikBernhardt:
The documentation of `IsPrimePowerInt` states:

> Note that IsPrimePowerInt uses SmallestRootInt (14.2-10) and a probable-primality test. 

Note that the mentioned probable-primality test is used in `IsPrimeInt`. Are we happy to accept this? If so I'm happy to approve this PR.

wilfwilson:
Looking at the GAP documentation, this seems to only apply for primes that are greater than 10^18, which is probably unlikely to appear.

Here's another implementation:
```
IsPowerOfPrime := function( n, p )
  if not IsPosInt(n) or n = 1 then
    return false;
  fi;
  return p ^ PVALUATION_INT(n, p) = n;
end;
```

DominikBernhardt:
I am happy for this to be merged.

fingolfin:
Some more timings:
```
gap> IsPowerOfPrimeNEW:={n,p} -> IsPosInt(n) and (n mod p = 0) and IsPrimePowerInt(n);;
gap> IsPowerOfPrimeNEW2:={n,p} -> IsPosInt(n) and (n mod p = 0) and p ^ PVALUATION_INT(n, p) = n;;
gap> IsPowerOfPrimeNEW3:={n,p} -> n > 1 and p ^ PVALUATION_INT(n, p) = n;;
gap> Number([1 .. 10000000], n -> IsPowerOfPrime(n, 2)); time;
23
15586
gap> Number([1 .. 10000000], n -> IsPowerOfPrimeNEW(n, 2)); time;
23
2506
gap> Number([1 .. 10000000], n -> IsPowerOfPrimeNEW2(n, 2)); time;
23
1953
gap> Number([1 .. 10000000], n -> IsPowerOfPrimeNEW3(n, 2)); time;
23
1556
```

Note that the main speed up in the last version is to replace `IsPosInt` by `n > 1`.

wilfwilson:
I felt a bit uneasy about using an undocumented GAP kernel function, but it seems spread throughout the GAP library, so I'll use it.

fingolfin:
Hmm, good point, I didn't think about it. Of course one could use `PValuation`, which is documented, but it's a GAP function and using it makes the code slower than what you originally proposed. 

Anyway, I am fine with any of these variants. In the end, if this ever turns out to be a bottleneck, we could produce much faster versions via a kernel extension. And even faster if we specialize to `p=2`. So, I am not worried about this. I do like the simpler code, though.

In short: please pick whichever version you personally like best, I'll be happy to go with it.

wilfwilson:
Then I'll stick with what I just pushed `IsPowerOfPrime := {n, p} -> n > 1 and p ^ PVALUATION_INT(n, p) = n;`

PR #144, opened 2019-07-10
Replace many "Error"s with "ErrorNoReturn"

wilfwilson:
I did this everywhere that I thought appropriate (in particular, I didn't change `misc/` or `contrib/`; I changed most occurrences in `gap/`).

An error is usually triggered when something wrong has happened. `Error` lets you return after this, and continue the computation, which is almost always a bad idea if something genuinely wrong has happened. `ErrorNoReturn` only lets you quit the computation after an error, which seems like the most sensible option in most cases.

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -121,7 +121,7 @@ InstallGlobalFunction( RecogniseGroup,
     elif IsMatrixGroup(G) then
         return RecogniseGeneric(G,FindHomDbMatrix,"");
     else
-        Error("Only matrix and permutation groups are supported");
+        ErrorNoReturn("Only matrix and permutation groups are supported");
     fi;
 
 # TODO: perhaps check if the result does not have IsReady set;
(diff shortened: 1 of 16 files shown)


wilfwilson:
Sure thing. Basically the question is: are there now `ErrorNoReturn`s in places that you *would* actually want to continue a computation?

gap/base/recognition.gi
```diff
@@ -1024,7 +1024,7 @@ RECOG.TestGroup := function(g,proj,size, optionlist...)
           Print("proj := ", proj, ";\n");
           Print("seedMT := ", seedMT, ";\n");
           Print("seedRS := ", seedRS, ";\n");
-          Error("Alarm: Size not correct!\n");
```
ChrisJefferson: this should not be errornoreturn

gap/base/recognition.gi
```diff
@@ -1055,7 +1055,7 @@ RECOG.TestGroup := function(g,proj,size, optionlist...)
           if s <> fail then
             Print("result := ", ResultOfStraightLineProgram(s,l), ";\n");
           fi;
-          Error("Alarm: SLPforElement did not work!\n");
```
ChrisJefferson: or this

gap/base/recognition.gi
```diff
@@ -1087,7 +1087,7 @@ RECOG.TestGroup := function(g,proj,size, optionlist...)
                 Print("result := ", ResultOfStraightLineProgram(s,l), ";\n");
             fi;
 
-            Error("Alarm: SLPforElement did not work on (possibly) non-group element!\n");
```
ChrisJefferson: or this

PR #147, opened 2019-07-11
Added dependencies information

hulpke:
Text and graphical version of function dependencies (as promised as part of documentation). Files currently are in misc, but easily could go somewhere else. The pdf version is created from the `.dot` version using the graphviz program suite.
A PDF version of the graph is at
[dependencies.pdf](https://github.com/gap-packages/recog/files/3381505/dependencies.pdf)


hulpke:
@fingolfin 
Thanks for the suggestion. The PDF was only there in case someone could not interpret the `dot` file. I will attach it to this PR instead.

PR #150, opened 2019-07-12
Fix documentation for and rename Perm.VeryFewPoints

### LINKED ISSUE #141: FindHomMethodsPerm.VeryFewPoints: very few, or very few and small?
From the documentation for `FindHomMethodsPerm.VeryFewPoints`:
> If a permutation group acts only on a few points (the current limit is at most 10 points) then a stabiliser chain is computed by the randomized GAP library function for that purpose.

This suggests to me that it makes a stabiliser chain for `G` if `NrMovedPoints(G) <= 10`. But it's actually only making a stabiliser chain if `IsSubset([1..10], MovedPoints(G))`, i.e.:
```
if LargestMovedPoint(G) <= 10 then
  return FindHomMethodsPerm.StabChain(ri, G);
fi;
return NeverApplicable;
```
Which behaviour do we want? In either case, I think we should update the documentation here, and possibly the name of this function, to make it crystal clear what is happening.

wilfwilson:
This resolves #141.

If people object to the renaming, then I don't mind dropping that commit.

### DIFF
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -16,7 +16,7 @@
 ##
 #############################################################################
 
-#! @BeginChunk VeryFewPoints
+#! @BeginChunk MovesOnlySmallPoints
 #! If a permutation group moves only small points
 #! (currently, this means that its largest moved point is at most 10),
 #! then this method computes a stabilizer chain for the group via
@@ -28,7 +28,7 @@
 #! If the input group moves a large point (currently, this means a point
 #! larger than 10), then this method returns <K>NeverApplicable</K>.
 #! @EndChunk
-FindHomMethodsPerm.VeryFewPoints := function(ri, G)
+FindHomMethodsPerm.MovesOnlySmallPoints := function(ri, G)
   if LargestMovedPoint(G) <= 10 then
       return FindHomMethodsPerm.StabChain(ri, G);
   fi;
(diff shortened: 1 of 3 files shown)


gap/perm.gi
```diff
@@ -16,19 +16,19 @@
 ##
 #############################################################################
 
-#! @BeginChunk VeryFewPoints
-#! If a permutation group acts only on a few points (the current limit is at
-#! most 10 points) then a stabiliser chain is computed by the randomized &GAP;
-#! library function for that purpose. If the method is successful then the
+#! @BeginChunk VerySmallPoints
```
fingolfin: Perhaps something like `MovesOnlySmallPoints` or a variation thereof is even more self-explaining?

gap/perm.gi
```diff
@@ -16,19 +16,19 @@
...
+#! If a permutation group moves only very small points
+#! (currently, this means that its largest moved point is at most 10),
+#! then a stabiliser chain is computed by the randomized &GAP;
+#! library function for that purpose. If the method is successful, then the
 #! calling node becomes a leaf node in the recursive scheme. If the input group
-#! acts on more than 10 points then the method returns <C>NeverApplicable</C>.
+#! acts on more than 10 points, then the method returns <C>NeverApplicable</C>.
```
fingolfin: I realize you only edited the text in a minor way. But perhaps we could also expand the explanation a bit: I think the reason for the existence of this method is that when dealing with very small degree permutation groups, there is little point in "reognition", as just computing a stab chain is faster than anything fancy we might do.

So perhaps we could mention this briefly?

Of course that's true if all we care about is "some" recognition tree. If we want to do more serious work with it, we may want to really "recognize" every part explicitly... Whether this means this method should be disabled in the future, or whether instead one should explicitly rewrite the initial recognition tree into a more cleverly arrange composition tree (as done in Magma) is another question.

Anyhow: I am also fine with merging this PR as-is!

gap/perm.gi
wilfwilson: I like it.

gap/perm.gi
wilfwilson: I had a go.

PR #152, opened 2019-07-12
Update Travis settings

fingolfin:
- switch from trusty to xenial for newer versions of everything
- don't bother testing a 32bit variant
- but do test against stable GAP branches in addition to master
- remove obsolete env variables

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ dist: xenial
 
 language: c
 
+env:
+  global:
+    - GAP_PKGS_TO_CLONE="AutoDoc"
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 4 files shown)


wilfwilson:
GAP package updates are not being picked up at the moment, I think Alexander is on holiday, so to get a new enough Autodoc you'll have to do it manually.

.travis.yml
```diff
@@ -15,23 +15,17 @@ addons:
...
+    - env: GAPBRANCH=stable-4.10
 
 before_script:
-  - export GAPROOT="$HOME/gap"
   - git clone https://github.com/gap-system/pkg-ci-scripts.git scripts
   - scripts/build_gap.sh
 script:
-  - scripts/build_pkg.sh && scripts/run_tests.sh
+  - scripts/build_pkg.sh
+  - $HOME/gap/bin/gap.sh -l "$PWD/gaproot;" --quitonbreak makedoc.g
```
wilfwilson: `makedoc.g` no longer quits GAP when it finishes. I'm not sure of a nice way of making GAP still quit, but this is one way:
```
bash echo "quit;" | $HOME/gap/bin/gap.sh -l "$PWD/gaproot;" --quitonbreak makedoc.g
```

PR #153, opened 2019-07-12
Documentation changes (mostly formatting)

wilfwilson:
This has four commits:
* Make sure that all booleans are tagged in documentation with `<K>` tags
* Tag the words `Success`, `TemporaryFailure`, `NotEnoughInformation`, and `TemporaryFailure` with `<K>` tags
* Change some <C>(function name)</C> things to references
* A tiny code change that I made in the spirit of making things shorter :wink:

The second of these is the only controversial one in my opinion: these variables are not GAP keywords, but I still like to use the `<K>` tags to emphasise their specialness, and because GAPDoc gives them highlights them in a way that I like.


gap/base/recognition.gi
```diff
@@ -871,10 +871,7 @@ end );
 InstallGlobalFunction( FindKernelFastNormalClosure,
   # Used in the generic recursive routine.
   function(ri,n1,n2)
-    local succ;
-
-    succ := FindKernelRandom(ri,n1);
-    if succ = false then
+    if not FindKernelRandom(ri,n1) then
```
```suggestion
    if not FindKernelRandom(ri, n1) then
```

PR #154, opened 2019-07-12
Refactor and add comments to base/projective.gi

wilfwilson:
Is this kind of thing welcome? I enjoy doing it anyway, even though arguably I could be doing something more productive.

I took a look at this file, and thought that I could both shorten the code and make it clearer what is happening.

I'm not sure whether the `Assert`s (and their levels) are appropriate.

### DIFF
--- a/gap/base/projective.gi
+++ b/gap/base/projective.gi
@@ -16,50 +16,53 @@
 ##
 #############################################################################
 
+# Assumes <m> is a non-empty square matrix defined over a field. <m> defines a
+# projective one iff it is a non-zero scalar multiple of the identity matrix.
 InstallGlobalFunction( IsOneProjective,
-  function(el)
-    local s, n, i, j, zero;
-    n := NrRows(el);
-    Assert(0, n = NrCols(el));
-    s := el[1,1];
-    if IsZero(s) then return false; fi;
-    zero := Zero(s);
-    for i in [1..n] do
-        if el[i,i] <> s then return false; fi;
-        for j in [1..n] do
-            if i <> j and el[i,j] <> zero then return false; fi;
-        od;
-    od;
-    return true;
+  function(m)
+    Assert(1, NrRows(m) > 0 and NrRows(m) = NrCols(m));
+    return not IsZero(m[1,1]) and RECOG.IsScalarMat(m) <> false;
   end );
 
+# Assumes <a> and <b> are non-empty equal-dimension square matrices over equal
+# fields, and that the first row of <a> contains a non-zero entry.
+#
+# <a> and <b> define equal projective elements, or are "equal modulo scalars",
+# iff there exists a (non-zero) scalar <s> in the field such that <s * a = b>.
 InstallGlobalFunction( IsEqualProjective,
-  function(a,b)
-    local p,s,i;
-    p := PositionNonZero(a[1]);
-    s := b[1,p];
-    if IsZero(s) then return false; fi;
-    s := s / a[1,p];
-    for i in [1..Length(a)] do
-        if s*a[i] <> b[i] then return false; fi;
-    od;
-    return true;
+  function(a, b)
+    local n, p, s;
+    n := NrRows(a);
+    Assert(1, n > 0 and n = NrRows(b) and n = NrCols(a) and n = NrCols(b));
+    p := First([1..n], i -> not IsZero(a[1,i])); # Find non-zero entry in <a[1]>
+    Assert(1, p <> fail);
+    s := b[1,p] / a[1,p]; # The unique scalar <s> with <s * a[1,p] = b[1,p]>.
+    return not IsZero(s)
+           and ForAll([1..n], i -> ForAll([1..n], j -> s * a[i,j] = b[i,j]));
   end );
 
-RECOG.ProjectiveOrder := function(el)
-  return ProjectiveOrder(el)[1];
-end;
+# The projective order of an invertible square matrix <m> over a finite field is
+# the least positive integer <k> such that <m ^ k> is a projective one.
+RECOG.ProjectiveOrder := m -> ProjectiveOrder(m)[1];
 
+# Assumes <m> is a non-empty square matrix over a field.
+#
+# <m> is a scalar matrix iff it is a scalar multiple of the identity matrix,
+# i.e. a diagonal square matrix with the same value along its main diagonal.
+#
+# If <m> is a scalar matrix, then this function returns the scalar that appears
+# on its main diagonal. Otherwise, this function returns <false>.
 RECOG.IsScalarMat := function(m)
-  local i,x;
-  if not(IsDiagonalMat(m)) then
-      return false;
-  fi;
+  local x;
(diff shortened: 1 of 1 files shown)


gap/base/projective.gi
```diff
@@ -16,50 +16,48 @@
...
-    s := b[1,p];
-    if IsZero(s) then return false; fi;
-    s := s / a[1,p];
-    for i in [1..Length(a)] do
-        if s*a[i] <> b[i] then return false; fi;
-    od;
-    return true;
+  function(a, b)
+    local p, s;
+    Assert(2, NrRows(a) = NrRows(b) and not IsEmpty(a));
```
fingolfin: `IsEmpty` is defined for collections, but MatrixObj are not generally collections (they don't support `Size` either, for example). Though I guess we might want to add something like `IsEmptyMatrix` or so to the MatrixObj API, with a default implementation that returns `NrCols(m)=0 or NrRows(m)=0`

gap/base/projective.gi
```diff
@@ -16,50 +16,48 @@
...
+    p := PositionNonZero(a[1]);  # Find a non-zero entry in first row of <a>.
+    s := b[1,p] / a[1,p];        # The unique <s> with <s * a[1,p] = b[1,p]>.
+    return not IsZero(s) and ForAll([1 .. NrRows(a)], i -> s * a[i] = b[i]);
```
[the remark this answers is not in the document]
wilfwilson: Thanks for that insight. I removed this usage, as well as the usage of `a[1]` two lines above.

gap/base/projective.gi
wilfwilson: Ah right. I changed these sorts of checks to `NrRows(a) > 0`.

gap/base/projective.gi
```diff
@@ -16,50 +16,53 @@
...
-    for i in [1..Length(a)] do
-        if s*a[i] <> b[i] then return false; fi;
-    od;
-    return true;
+  function(a, b)
+    local n, p, s;
+    n := NrRows(a);
+    Assert(1, n > 0 and n = NrRows(b) and n = NrCols(a) and n = NrCols(b));
+    p := First([1..n], i -> not IsZero(a[1,i])); # Find non-zero entry in <a[1]>
+    Assert(1, p <> fail and BaseDomain(a) = BaseDomain(b));
```
wilfwilson: @fingolfin Might the `BaseDomain` check here be horribly slow?

gap/base/projective.gi
```diff
@@ -16,50 +16,53 @@
...
+    s := b[1,p] / a[1,p]; # The unique scalar <s> with <s * a[1,p] = b[1,p]>.
+    return not IsZero(s)
+           and ForAll([1..n], i -> ForAll([1..n], j -> s * a[i,j] = b[i,j]));
```
fingolfin: Note that using ForAll like this is a bit slower than using an explicit unrolled loop; this is mostly due to the overhead for calling a GAP function for every single matrix entry (in the worst case).
And, sadly, switching from row-based to entry based checks can *totally* kill performance.

 We can verify all this with a micro benchmark:
```
gap> n:=100;; a:=IdentityMat(n, GF(3));;
gap> s:=Z(3)^0;;
gap>
gap> a:=a*s;; # uncompressed FFE matrix
gap> for i in [1..1000] do for i in [1..n] do for j in [1..n] do x := a[i,j] = s*a[i,j]; od; od; od; time;
1696
gap> for i in [1..1000] do ForAll([1..n], i -> ForAll([1..n], j -> a[i,j] = s*a[i,j])); od; time;
1940
gap> for i in [1..1000] do for i in [1..n] do x := a[i] = s*a[i]; od; od; time;
36
gap> for i in [1..1000] do ForAll([1..n], i -> a[i] = s*a[i]); od; time;
36
gap> for i in [1..1000] do x := a = s*a; od; time;
20
gap>
gap> ConvertToMatrixRep(a);; # compressed matrix
gap> for i in [1..1000] do for i in [1..n] do for j in [1..n] do x := a[i,j] = s*a[i,j]; od; od; od; time;
1692
gap> for i in [1..1000] do ForAll([1..n], i -> ForAll([1..n], j -> a[i,j] = s*a[i,j])); od; time;
2048
gap> for i in [1..1000] do for i in [1..n] do x := a[i] = s*a[i]; od; od; time;
32
gap> for i in [1..1000] do ForAll([1..n], i -> a[i] = s*a[i]); od; time;
48
gap> for i in [1..1000] do x := a = s*a; od; time;
80
```
I therefore suggest that we check for `IsRowListMatrix`, and try to fallback to the row based approach if possible; and if that is not possible, then we use `s*a = b` instead (which needs to create a complete new matrix, which causes the extra overhead seen in the last test case above).

For optimal performance, one really needs to taylor the code to the matrix representation. That would suggest turning this (and the other function in this file) into operations, allowing the installation of additional methods for other IsMatrixObj implementations. *Except* that this will slow down things for "classic" `IsMatrix` plist-of-plist matrices, which trigger method selection over head... But arguably, we shouldn't be seeing these here anyway, only compressed FFE matrices... So perhaps this operation approach would be the cleanest way to achieve what I suggest above (and we could even have a method which triggers for plist inputs, and prints out a performance warning)

gap/base/projective.gi
```diff
@@ -16,50 +16,53 @@
...
-  fi;
+  local x;
+  Assert(1, NrRows(m) > 0 and NrRows(m) = NrCols(m));
   x := m[1,1];
-  for i in [2..Length(m)] do
-      if m[i,i] <> x then
-          return false;
-      fi;
-  od;
+  if ForAny([2..NrRows(m)], i -> m[i,i] <> x) or not IsDiagonalMat(m) then
```
fingolfin: Nice thing about this is that `IsDiagonalMat` may be overloaded with a faster method for specific MatrixObj implementations, so this may be a win in performance.

gap/base/projective.gi
```diff
@@ -4,7 +4,7 @@
 ##  which provides a collection of methods for the constructive recognition
 ##  of groups.
 ##
-##  This files's authors include Max Neunhöffer, Ákos Seress.
+##  This files's authors include Max Neunhöffer, Ákos Seress, Wilf Wilson.
```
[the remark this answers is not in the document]
wilfwilson: Thanks for the food ;-)

gap/base/projective.gi
wilfwilson: Alright, I removed the check from the `Assert`.

gap/base/projective.gi
wilfwilson: I've pushed a second commit taking on board some of what you've said. If it's what you meant, then I can squash it into the main commit, if not I can make some changes.

I agree that using operations should be the best idea for optimal performance, but I don't want to do it with this PR. I've done as much work on this file as I want to at the moment, and since (at least I feel) this  PR improves things, I think it would be nice to get it merged before I/you/someone else makes future additional improvements 🙂

PR #155, opened 2019-07-14
Tweak and reformat code in several files

wilfwilson:
I looked through several files one evening last week, and for fun I started to make some mostly superficial changes as I went along. The idea with this PR is to go some small way to making the code in the package more readable, and also more consistent with what I know of the GAP world. Sorry if this is overwhelming to review.

Here are the main things that I've done:
* I broke one-line conditions `if condition then something; fi;` into three-liners.
* I removed brackets around simple negations, i.e. `not(single_condition)` -> `not single_condition`.
* I joined some consecutive and compatible `if ... fi; if ... fi;` bits of code with `elif`s.
* I used `StringFormatted` to clean up some messy `Info` messages.
* I slightly improved `RECOG.IsDiagonalBlockOfMatrix` in `gap/matrix.gi`.
* I corrected a comment in `RECOG.InvolutionCentraliser` in `gap/d247.gi`.
* I changed `return false;` to `return NeverApplicable;` in `FindHomMethodsProjective.Subfield`.
* I replaced some `Set(List(...` calls with `Set(...`.
* ~~I replaced a `List[Length(list)]` with `Last(list)`.~~
* I simplified some code in other ways, where I saw an opportunity to do so.

I also changed these somewhat more subjective things, simply because I wanted to:
* I made some record definitions line up nicely.
* ~~`Length(variable) = 0` to `IsEmpty(variable)`.~~
* ~~`Length(variable) > 0` to `not IsEmpty(variable)`.~~
* A few other spacing tweaks (sorry)
* I removed some final `else`s from the end of functions, i.e.
```gap
if condition then
  return value;
else
  return other;
fi;
```
becomes
```gap
if condition then
  return value;
fi;
return other;
```

### DIFF
--- a/gap/almostsimple.gi
+++ b/gap/almostsimple.gi
@@ -779,7 +779,9 @@ FindHomMethodsProjective.ThreeLargeElOrders := function(ri,G)
           fi;
           res := LookupHintForSimple(ri,G,namecat);
       fi;
-      if res = true then return Success; fi;
+      if res = true then
+          return Success;
+      fi;
   od;
   Info(InfoRecog,2,"Did not succeed with hints, giving up...");
   return fail; # FIXME: fail = TemporaryFailure here really correct?
@@ -919,7 +921,9 @@ end;
 RECOG.HomFDPM := function(data,x)
   local r;
   r := RECOG.FindPermutation(data.cob*x*data.cobi,data.fdpm);
-  if r = fail then return fail; fi;
+  if r = fail then
+      return fail;
+  fi;
   return r[2];
 end;
 
@@ -1279,7 +1283,9 @@ RECOG.RuleOutSmallProjOrder := function(m)
   local l,o,v;
   if IsPerm(m) then
       o := Order(m);
-      if o > 119 then return fail; fi;
+      if o > 119 then
+          return fail;
+      fi;
       return o;
   fi;
   v := ShallowCopy(m[1]);
@@ -1287,15 +1293,21 @@ RECOG.RuleOutSmallProjOrder := function(m)
   ORB_NormalizeVector(v);
   o := Orb([m],v,OnLines,rec( hashlen := 300, report := 0 ));
   Enumerate(o,121);
-  if not(IsClosed(o)) then return fail; fi;
+  if not IsClosed(o) then
+      return fail;
+  fi;
   l := Length(o);
   Randomize(v);
   ORB_NormalizeVector(v);
   o := Orb([m],v,OnLines,rec( hashlen := 300, report := 0 ));
   Enumerate(o,121);
-  if not(IsClosed(o)) then return fail; fi;
+  if not IsClosed(o) then
+      return fail;
+  fi;
   l := Lcm(l,Length(o));
-  if l > 119 then return fail; fi;
+  if l > 119 then
+      return fail;
+  fi;
   return l;
 end;
 
--- a/gap/almostsimple.gi
+++ b/gap/almostsimple.gi
@@ -21,12 +21,10 @@
 RECOG.ParseNumber := function( number, d, default )
   if IsInt(number) then
       return number;
-  fi;
-  if IsString(number) then
-      if number = "logd" then return LogInt(d,2); fi;
-      if number[Length(number)] = 'd' then
-          return d * Int(number{[1..Length(number)-1]});
-      fi;
+  elif number = "logd" then
+      return LogInt(d, 2);
+  elif IsString(number) and EndsWith(number, "d") then
+      return d * Int(number{[1..Length(number)-1]});
   fi;
   return default;
 end;
@@ -1341,8 +1339,7 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
       if o = fail then
           Info(InfoRecog,2,"Ruled out all sporadic groups.");
           return NeverApplicable;
-      fi;
-      if i <= Length(gens) then
+      elif i <= Length(gens) then
           r.order := ri!.order(r.el);
       else
           GetElmOrd(ri,r);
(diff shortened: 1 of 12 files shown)


fingolfin:
So in general, I am in factor of such changes, but having them all lumped together in one giant commit makes it indeed hard to review, and also harder to pinpoint regressions later on via `git bisect`. It'd be far nicer if these were split into multiple commits, say grouped by the type of change. 

You also have to be careful, at least one of these changes (introducing use of `Last`) will break compatibility with GAP 4.10, which I'd like to avoid for now; requiring people to use a development version of GAP will just make it even harder to get people to help with recog. (I realize this was likely not intentional! Just wanted to point it out. Of course I am in general *for* the use of `Last`, I mean, I added it for just this :-). Anyway, it's good that I recently updated Travis to test against `stable-4.10`, too ;-)

fingolfin:
Another note: `Length(variable) = 0` tends to me a lot faster than `IsEmpty(variable)` because it is a specialized kernel operation, and avoids method dispatch overhead when given a plist as argument. So esp. if the input is a list of (old style "list of list") matrices, the extra overhead can accumulate quickly.

wilfwilson:
Thanks for the review @ChrisJefferson.

I'm glad you're in favour in general, @fingolfin. I appreciate that smaller commits make reviewing and bisecting nicer, so I've gone and done that. Breaking compatibility with GAP `stable-4.10` was indeed unintentional, and since I was aware that Travis now tests this branch, I assumed that Travis would catch any problems, but obviously not.

I've removed the `Last` (and added a comment about a future possible change) and removed the newly-introduced calls to `IsEmpty`.

gap/almostsimple.gi
```diff
@@ -21,10 +21,10 @@
...
   if IsInt(number) then
       return number;
-  fi;
-  if IsString(number) then
-      if number = "logd" then return LogInt(d,2); fi;
-      if number[Length(number)] = 'd' then
+  elif IsString(number) then
+      if number = "logd" then
+          return LogInt(d, 2);
+      elif number[Length(number)] = 'd' then  # Can eventually be Last(number)
```
fingolfin: You could further compact this to:  `if IsInt(number) ... elif number = "logd" ... elif IsString(number) and EndsWith(number, "d") ... fi;` (this also avoids the use of `Last`)

fingolfin:
@wilfwilson huh, actually, we *do* test against `stable-4.10` on Travis now. So I am not sure how the tests could have passed while you were using `Last`...

Aha, looking at the build logs, I see this:
```
Syntax warning: Unbound global variable in /home/travis/build/gap-packages/rec\
og/gaproot/pkg/recog/gap/almostsimple.gi:27
      elif Last(number) = 'd' then
               ^
```

Hrm, perhaps we need an option "abort on syntax error" for GAP... Also, I think this shows that this part of the function is never ever called by anything, in particular no tests, else it would have triggered a break loop. That is a bit concerning in itself...

gap/classical.gi
```diff
@@ -581,13 +585,14 @@ IsPrimitivePrimeDivisor := function( b, a, p )
...
 
-    if not IsPrimeInt(p) then return false; fi;
-    if not IsInt(b) or b <= 1 then return false; fi;
-    if not IsInt(a) or a <= 1 then return false; fi;
-
-    if (b^a-1) mod p <> 0 then return false; fi;
+    if not (IsPrimeInt(p) and IsInt(a) and IsInt(b) and a > 1 and b > 1 and
+            b^a-1 mod p = 0) then
+        return false;
+    fi;
```
fingolfin: Actually this is one of the few places where I kind of disagree: Separating the four checks for me makes it easier to follow the code, as they have a kind of logical grouping: First we validate, p; then b, then a. After this, we start checking the actual conditions we are supposed to check.

In an unrelated matter, I think that if `IsPrimeInt(p)` is false, then we actually should call `Error`, because that indicates a usage error, and should not be silently swallowed. And I am tempted to say the say should hold for the initial checks of a and b.

gap/matrix.gi
```diff
@@ -795,13 +796,14 @@ SLPforElementFuncsMatrix.LowerLeftPGroup := function(ri,g)
...
-      if not(IsZero(h)) then return fail; fi;
+      if not IsZero(h) then
+          return fail;
+      fi;
   od;
   if Length(l) = 0 then
-      return StraightLineProgramNC([[1,0]],Length(ri!.gensNvectors));
-  else
-      return StraightLineProgramNC([l],Length(ri!.gensNvectors));
+      return StraightLineProgramNC([[1, 0]], Length(ri!.gensNvectors));
```
fingolfin: Could further simplify here:
```suggestion
      l := [1, 0];
```

gap/c3c5.gi
```diff
@@ -397,7 +397,7 @@ FindHomMethodsProjective.Subfield :=  function(ri,G)
     RECOG.SetPseudoRandomStamp(G,"Subfield");
     f := ri!.field;
     if IsPrimeField(f) then
-        return false;     # nothing to do
+        return NeverApplicable;     # nothing to do
```
fingolfin: Ooops, good catch! I don't understand how I missed that one. Small consolation for me is that you also missed two other `return` statements in this function ;-)

gap/almostsimple.gi
wilfwilson: Ah, nice suggestion.

gap/classical.gi
wilfwilson: I reverted this. In my opinion, there's a lot about `classical.gi` that could/should be cleaned up (I already started to have a go at this in a separate branch). It would probably be better to make changes of the kind I made here in a more targeted, coherent manner.

gap/c3c5.gi
wilfwilson: Well, I didn't miss them, I saw them but decided I didn't want to spare the thought to work out whether they _should actually_ also be `NeverApplicable`s. But it turns out it only took a moment's thought to realise that, yes, they should.

wilfwilson:
Thanks for the review @fingolfin. I made the changes you suggested.

Yes, I realised why Travis failed despite using `Last`: it must have only raised syntax warning, and that the code was never actually run. In the Semigroups and Digraphs packages CI setup, we save the output of GAP to a fail, and then grep it afterwards for various bad words and phrases, such as `Syntax warning` I think, and fail if so. It's not particularly elegant, but it works.

PR #158, opened 2019-07-15
Improve referencing in documentation in some parts of perm.gi

wilfwilson:
In #131 @PaulaHaehndel and co-authors added things like `<C>FindHomMethodsPerm.Imprimitive</C>` to certain parts of the autodoc documentation. I came along and blithely changed this to `<Ref Subsect="Imprimitive" Style="Text"/>` and similar, dropping the `FindHomMethodsPerm` bit. But I have now learnt that it was a deliberate choice to include in #131 (see https://github.com/gap-packages/recog/pull/131#discussion_r301122826)

I've tried to find a compromise. How do we feel about this? It renders fine in the manual, and I think looks fine in the code too. If we like it, I can do this in other appropriate places of the documentation.

### DIFF
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -120,12 +120,12 @@ FindHomMethodsPerm.Imprimitive :=
 
 #! @BeginChunk PcgsForBlocks
 #! This method is called after a hint is set in
-#! <Ref Subsect="Imprimitive" Style="Text"/>.
+#! <C>FindHomMethodsPerm.</C><Ref Subsect="Imprimitive" Style="Text"/>.
 #! Therefore, the group <A>G</A> preserves a non-trivial block system.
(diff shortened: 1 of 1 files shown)


fingolfin:
Two points:
- What is the advantage of seeing `FindHomMethodsPerm.Imprimitive` over just `Imprimitive`? The answer affects which alternatives make sense
- If we really want the full string, we could also change the generate documentation to include the full name as label, so that one could use `<Ref Subsect="FindHomMethodsPerm.Imprimitive" ...`

PaulaHaehndel:
I like to read `FindHomMethodsPerm` in the code. Especially in the example of `BlockScalarProj` (in projective) that references `BlockScalar` (in matrix). Where it is nice to have a hint where to search for it, while you are in the code.

fingolfin (approved):

PR #159, opened 2019-07-15
Change more `not(foo)` to `foo`; use `PrimeDivisors` in a few spots

fingolfin:


gap/matrix/ppd.gi
```diff
@@ -179,7 +179,7 @@ end;
 #             # This removes all irreducible factors of degree e from c
 #             # The product of those that divided c is gcd
 #             for div in DivisorsInt(e) do
-#               if not(div in result) then
+#               if notdiv in result then
```
wilfwilson:
```suggestion
#               if not div in result then
```

PR #175, opened 2020-06-17
Extend AddMethod

### LINKED ISSUE #107: Systematically verify that all SLPForElement functions and Homoms check input
Without that, we may end up with incorrect recognition of certain factors/images; with bad answers to user inputs in general; and also with unexpected `Error()`s during recognition.

Partial progress towards this was made in the past, see e.g. PRs #96, #102, and commits 231ac0210913fb39a97ce49011f5555b5097af98, 8a15287247385b26e1339f557c561f4b38bc5575

I suggest that we seek out each of the relevant functions, and inspect them; and in order to be able to conclude that we are done, and to avoid duplicate work, let's annotate each single spot we inspected with a comment, like this: `# Verified by Max Horn on 4 July 2019 to validate inputs` or so. Starting by annotating the changes made in the PRs and commits linked above.

Thoughts?

ssiccha:
Change AddMethod to now take two arguments: the method database and a record which contains the method to be added and information about it. 

Setting the InfoClass InfoAddMethod to 1 now produces messages when AddMethod is called with unbound method.validatesInput component.

I added a few small changes to the documentation. The PR is now ready for review.

Now we can address #107 method for method.

### DIFF
--- a/gap/base/methsel.gi
+++ b/gap/base/methsel.gi
@@ -30,6 +30,12 @@ InstallGlobalFunction("AddMethod", function(methodDb, method)
     if not IsBound(method.comment) then
         method.comment := "";
     fi;
+    # method.validatesOrAlwaysValidInput for now only stores meta-information.
+    # It may be used in the future, see github issue #184.
+    if not IsBound(method.validatesOrAlwaysValidInput) then
+        Info(InfoAddMethod, 1, "AddMethod, ", method.stamp,
+             ": validatesOrAlwaysValidInput not bound");
+    fi;
     l := Length(methodDb);
     pos := First([1 .. l], i -> methodDb[i].rank <= method.rank);
     if pos = fail then
--- a/gap/base/methsel.gi
+++ b/gap/base/methsel.gi
@@ -20,28 +20,22 @@
 # Add a method to a database with "AddMethod" and call a method from a
 # database with "CallMethods".
 #
-InstallGlobalFunction( "AddMethod", function(arg)
-  # First argument is the method database, second is the method itself,
-  # third is the rank, fourth is the stamp. An optional 5th argument is
-  # the comment.
-  local comment,db,i,l,mr,p;
-  if Length(arg) < 4 or Length(arg) > 5 then
-      ErrorNoReturn("Usage: AddMethod(database,method,rank,stamp [,comment] );");
-  fi;
-  db := arg[1];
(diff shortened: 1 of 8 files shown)


wilfwilson:
How is the `method.validatesInput` going to be useful? How are you going to use it? (Is it going to be documented, etc?)

wilfwilson (commented):
I'm fine with `AddMethod` taking a record rather than various arguments. I'm just not sure what you intend for the `validatesInput` component, which I asked about above.

ssiccha:
> How is the `method.validatesInput` going to be useful? How are you going to use it? (Is it going to be documented, etc?)

I'm going to use it to work on #107. E.g. for each homomorphism I'm going to make it explicit, whether it checks its input or not. For now I'd rather see this as a sort of meta-information on the homomorphism. We could later require in `AddMethod` the method to have a `method.validatesInput`, to make sure that the creator of the method did not simply forget about the input validation, as it apparently has happened with some methods already.

ssiccha:
If we were to go with requiring it in `AddMethod` we could document that in the "what is a method" section.

wilfwilson (approved):

ssiccha:
> For now I'd rather see this as a sort of meta-information on the homomorphism. 

I'll put this information into the PR.

wilfwilson:
Good plan.

ssiccha:
Thanks for the review! :)

ssiccha:
Small update: I renamed `validatesInput` to `validatesOrAlwaysValidInput`. I think there are some "homomorphisms" which do not need to check their input. For example the plain StabChain method shouldn't have to. I'll double-check though.

Another small change. I added some commas in the commit `base: AddMethod input now is database and a record`.

fingolfin (approved):

gap/base/methsel.gi
```diff
@@ -20,28 +20,28 @@
...
+    local l, pos;
+    if not IsSubsetSet(NamesOfComponents(method),
+                       ["method", "rank", "stamp"]) then
+        ErrorNoReturn("<method> must have components",
+                      " method, rank, and stamp");
+    fi;
+    if not IsBound(method.comment) then
+        method.comment := "";
+    fi;
+    # method.validatesOrAlwaysValidInput for now only stores meta-information.
```
fingolfin, line 33: I do find `validatesOrAlwaysValidInput` rather clunky, but since it's not yet used, I'll not object to it here; it'd be trivial to rename later on anyway.

Issue #17, opened 2016-08-04
gap/SnAnBB.gi is unused, what to do with it?


fingolfin:
I discovered that the code in `gap/SnAnBB.gi` is completly unused -- that is, anything calling it from outside (namely from `gap/almostsimple.gi`) is itself dead. I added various FIXMEs.

Also, some of that code is duplicated in `gap/giant.gi`.

The question is: Is this obsolete code we should throw away? Or is that "new" code which Akos and Max were working on integrating, but simply never finialized it?

fingolfin:
I should point out that `SnAnBB.gi` says this:
```
##  The code is based upon the algorithm presented in the paper
##       "A black-box group algorithm for recognizing 
##        finite symmetric and alternating groups. I." 
##  by Robert Beals, Charles R. Leedham-Green, Alice C. Niemeyer,
##     Cheryl E. Praeger and 'Akos Seress.
```

So hopefully Alice can comment on whether we should keep this code or not...

ssiccha:
I have also got martin's new code and will try to work on this.

ssiccha:
The reason why `SnAnBB.gi` probably was not used is that the code in `almostsimple.gi:933` only passes degrees to `SnAnBB.gi` for the action on the deleted permutation module.
Once we plug in the code from Jambor et al. which determines the degree of Sn or An everything wrt to Sn and An should run smoothly.

fingolfin:
The current code in `SnAnBB.gi` is completely superseded by the Jambor et. al. code in PR #42. So there is no point in trying to hook it up now.

ssiccha:
I agree.

SoongNoonien:
The code in this file has changed over the years and is definitely used now.

SoongNoonien:
See also #265 for more recent changes to the code using `gap/SnAnBB.gi`.

PR #176, opened 2020-06-18
Implement "recognize An/Sn of unknown degree" algorithm by Jambor et al.

### LINKED ISSUE #29: Add algorithm to recognize An and Sn on non deleted permutation modules
The method [FindHomMethodsProjective.AltSymBBByDegree](https://github.com/gap-packages/recog/blob/c67974645e6adfa1befa93a8959928f3502c1fe3/gap/almostsimple.gi#L895) uses an inefficient method to do BB recognition of Sn and An on modules that are not deleted permutation modules.

Instead the degree of the permutation should be determined by using an algorithm by Jambor et al.
The paper is "Jambor, Sebastian, Leuner, Martin, Niemeyer, Alice C., PLESKEN, W.
Fast recognition of alternating groups of unknown degree. J. Algebra 392 (2013), 315-335."
For an implementation see https://github.com/aniemeyer/recogalmostsimple/ in the subfolder `to-add/AnSn`

### LINKED ISSUE #174: Implement "Fast recognition of alternating groups of unknown degree"
We need an implementation of the algorithm in "Fast recognition of alternating groups of unknown degree" by Jambor, Leuner, Niemeyer and Plesken. This should in the end cover symmetric groups, too.

There is an existing GAP implementation, but it's not in a great state, so it's probably easier to rewrite this from scratch, and only occasionally looking at the old code (which one can find in the close PR #42).

@DominikBernhardt once was interested in this, but I think he's busy with other things, like working on his actual phd project ;-).

ssiccha:
This is work in progress by @FriedrichRober and me.

Fixes #174.
Fixes  #29.


gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+    end;
+    cg := c ^ g;
+    cg2 := cg ^ g;
+    cg3 := cg2 ^ g;
+    cg4 := cg3 ^ g;
+    H1 := constructH1(c, cg, cg3, cg4);
+    # Test whether an elm of the set X, here called XX, commutes with at least
+    # two elements of H1.
+    XX := ListWithIdenticalEntries(3, 0);
+    XX[1] := c ^ r;
```
fingolfin, line 180: Instead of allocating a list here, just use three local vars? E.g x1, x2, x3.

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+        f2,
+        # smallest non-fixed point
+        m,
+        # integer, loop variable over [1 .. k]
+        j,
+        # element of G, loop variable
+        t,
+        # conjugating element
+        x;
+    F := ListWithIdenticalEntries(4,false);
```
fingolfin, line 220: More idiomatic:
```suggestion
    F := [false, false, false, false];
```

FriedrichRober: Done

gap/SnAnUnknownDegree.gi
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+        local H1;
+        H1 := ListWithIdenticalEntries(5,0);
+        H1[1] := c ^ 2;
+        H1[2] := c ^ cg;
+        H1[3] := H1[2] ^ cg3;
+        H1[4] := H1[3] ^ cg3;
+        H1[5] := H1[4] ^ cg4;
+        return H1;
+    end;
+    constructH2 := function(c, cg, cg3, cg4)
```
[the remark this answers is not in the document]
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+        # respectively c ^ (g ^ i)
+        cg, cg2, cg3, cg4,
+        # temporary holder of H1, H2
+        temp,
+        # sets of elements of G
+        H1, H2, XX,
+        # helper functions
+        constructH1, constructH2, isElmPassingTest;
+    # Helper functions
+    constructH1 := function(c, cg, cg3, cg4)
```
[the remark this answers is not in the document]
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+        local H1;
+        H1 := ListWithIdenticalEntries(5,0);
+        H1[1] := c ^ 2;
+        H1[2] := c ^ cg;
+        H1[3] := H1[2] ^ cg3;
+        H1[4] := H1[3] ^ cg3;
+        H1[5] := H1[4] ^ cg4;
+        return H1;
```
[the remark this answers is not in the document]
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
+# Input: Group G, upper error bound eps, upper degree bound N
+#
+# The following algorithm constructs a set of possible 3-cycles. It is based
```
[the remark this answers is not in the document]
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+# on the simple observation that the product of two involutions t1, t2, which
+# only move one common point, squares to a 3-cycle.
+#
+# Either returns a list of elements of G or NeverApplicable
+BindGlobal("ThreeCycleCandidates",
+function(G, eps, N, groupIsOne, groupIsEq)
```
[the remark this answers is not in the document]
FriedrichRober: We will change this at the end. For now we leave it as it is so we can test the functions more easily in the console.

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,295 @@
...
+    return result;
+end);
+
+# g: a cycle matching c of a group G
+# c: a 3-cycle of a group G
+# r: arbitrary element of a group G
+# The supports of c and c^(g^2) have exactly one point, say alpha, in common.
+# Let phi be an isomorphism from G to a natural alternating or symmetric group.
+# This function decides whether alpha is a fixed point of phi(r).
+BindGlobal("IsFixedPoint",
```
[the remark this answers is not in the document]
ssiccha: Will do later, yes. :)

gap/SnAnUnknownDegree.gi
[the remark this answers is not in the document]
ssiccha: Actually, we'll try to do this next. Sounds good. :)

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,312 @@
...
+    t := fail;
+
+    # Constants
+    # TODO: better iteration over primes
+    M := 1;
+    p := 3;
+    while p <= N do
+        M := M * p ^ LogInt(N, p);
+        p := NextPrimeInt(p);
+    od;
```
fingolfin, line 33: How about adding and using this helper function (based on a code snippet by @rbehrends )
```
AllPrimesUpTo := function(n)
    local i, j, sieve, result;
    if n <= 1000 then
      return Primes{[1..PositionSorted(Primes, n+1)-1]};
    fi;
    sieve := BlistList([1..n], [1..n]);
    sieve[1] := false;
    for i in [2..Int(n/2)] do
        sieve[i*2] := false;
    od;
    i := 3;
    while i * i <= n do
        if sieve[i] then
            j := 3*i;
            while j <= n do
                sieve[j] := false;
                j := j + 2*i;
            od;
        fi;
        i := i + 2;
    od;
    return ListBlist([1..n], sieve);
end;
```

gap/SnAnUnknownDegree.gi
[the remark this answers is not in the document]
fingolfin: If it is difficult to test it via the console when using a recog info record, then we should make it easier to test with a proper recog info record (e.g. by adding helpers to produce them for you). You might in fact want to think about creating such helpers. Because everybody developing a recog method needs to eventually use those, and the sooner, the better (you avoid the time to rewrite the code from one API to another, and the risk of subtle mismatches). Also, all unit tests need that.

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+    R := Int(Ceil(7 / 4 * Log2(Float(eps ^ -1))));
+    S := 7 * N * R;
+    prebolsteringElms := [];
+    i := 0;
+    # find pre-bolstering elements
+    while i <= S and Length(prebolsteringElms) <= R do
+        r := RandomElm(ri,"simplesocle",true)!.el;
+        # test whether r is pre-bolstering
+        cr := c ^ r;
+        cr2 := c ^ (r ^ 2);
```
fingolfin, line 138: Perhaps better to write it like this (we avoid one squaring)
```suggestion
        cr2 := cr ^ r;
```

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+        if not isone(ri)(Comm(cr, c))
```
fingolfin, line 139: Computing a commutator is expensive for e.g. matrices where it requires two inversions and three multiplications. For permutations, it still requires one inversion, one conjugation and one multiplication. But we can get away with two multiplications here:
```suggestion
        if not isequal(ri)(cr*c, c*cr)
```
Of course you are not yet tuning for performance, but I thought I could just mention this thoughts while I had them :-).

One might be able to do even better for specific groups; and also, strictly speaking, it isn't necessary to produce temporary objects for the two products, one could write optimized code which doesn't need to produce the full two products as intermediate objects...

So perhaps we should have an `AreCommuting(ri)(g,h)` operation (as part of `ri`, to deal with projectives *sigh*) which does the "best" depending on the inputs?

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+                and not isequal(ri)(cr2, c)
+                and not isequal(ri)(cr2, c ^ 2)
+                and isone(ri)(Comm(cr2, c))
```
fingolfin, line 142: Sameremark as above

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+        then
+            Add(prebolsteringElms, r);
+        fi;
+        i := i + 1;
+    od;
+    # construct bolstering elements
+    for r in prebolsteringElms do
+        if isone(ri)((c ^ (r * c * r)
+                      * c ^ (r * c ^ (r ^ 2) * c)) ^ 3)
```
fingolfin, line 151: I think it would more sense to merge this into the preceeding loop: instead of sticking `r` into `prebolsteringElms` and then getting it out again, computing some more, and sticking the results into `result`; just move the processing code below up and replace the `Add(prebolsteringElms, r);` code line by it (and then adjust the condition of the surrounding loop, and remove `prebolsteringElms`.

What you gained by this is that you have `cr` and `cr2` available
```suggestion
        if isone(ri)((cr ^ (c * r)
                      * cr ^ (cr2 * c)) ^ 3)
```

Of course in the end, we need to produce SLPs anyway, and perhaps they are clever enough to figure out some of this, but I am not sure I'd want to rely on that?

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+        # (sets of) elements of G
+        H1, H2, x1, x2, x3,
+        # helper function
+        isElmPassingTest;
+    # Helper function
+    isElmPassingTest := function(ri, x, H)
+        local nrTrivialComm, h;
+        nrTrivialComm := 0;
+        for h in H do
+            if isone(ri)(Comm(x, h)) then
```
fingolfin, line 183: Another place `AreCommuting` (or `IsTrivialComm` or...) could be employed :-)

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,345 @@
...
+
+BindGlobal("AppendPoints",
+function(ri, g, c, r, s, k, k0)
+    local gc2, x, j;
+    gc2 := g * c ^ 2;
+    x := c;
+    # Invariant: x = c ^ (r ^ j)
+    for j in [1 .. k0] do
+        x := x ^ r;
+        if isone(ri)(Comm(x, gc2)) then
```
fingolfin, line 335: And another place for `AreCommuting` (and I bet there are more in recog)

gap/SnAnUnknownDegree.gi
```diff
@@ -481,4 +481,144 @@ function(ri, c, x, N)
...
+    local g, k, tmp, B, x;
+    B := BolsteringElements(ri, c, Float(eps) / 2., N);
+    if Length(B) < Int(Ceil(7./4. * Log(2. / Float(eps)))) then
+        return fail;
+    fi;
+    k := 0;
+    for x in B do
+        tmp := BuildCycle(ri, c, x, N);
+        if tmp = fail then
+            return fail;
```
ssiccha, line 504: As above this also should be TemporaryFailure?

gap/SnAnUnknownDegree.gi
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
FriedrichRober: Done

gap/SnAnUnknownDegree.gi
FriedrichRober: This is a very good idea. We try to add a framework for this operation.

gap/SnAnUnknownDegree.gi
ssiccha: Done.

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
 ##  The code is based upon the algorithm presented in the paper [BLGN+03].
 ##
 ##
-# DeclareInfoClass( "InfoRecSnAn" );
```
fingolfin, line 21: Please do not overwrite the existing BB code -- add yours to a new file. So that one can compare the two side by side, benchmark them against each other etc.

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
-# # However, it seems to me this whole file is dead code anyway..
-# RecSnAnIsOne := IsOne;
-# RecSnAnEq := EQ;
-#
-# ## is <r,s> in S_n
-#
-# SatisfiesSnPresentation := function( n, r, s )
-#
-#     local j, t;
+BindGlobal("SatisfiesSnPresentation",
```
fingolfin, line 21: This is not (yet?) used?

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
+        if not isone(ri)(Comm(s,t)^2) then
+            return false;
+        fi;
+        j := j + 1;
+    od;
+    return true;
+end);
+
+BindGlobal("SatisfiesAnPresentation",
+function(ri, s, t, n)
```
fingolfin, line 40: A bit confusing that `n` is in a different position here than for `SatisfiesSnPresentation`

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
+    else
+        # we already know s^(n-2) = t^3 = 1
+        if not(isone(ri)((s * t)^(n-1))) then
+            return false;
+        fi;
+        j := 1;
+        r := s^0;
+        while j <= (n-2)/2  do
+            r := r * s;
+            ti := t^-1;
```
fingolfin, line 71: Pull the computation of `ti` out of that loop?

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
+
+BindGlobal("SatisfiesAnPresentation",
+function(ri, s, t, n)
+    local j, r, ti;
+
+    if not(isone(ri)(s^(n-2))) or not(isone(ri)(t^3)) then
+        return false;
+    fi;
+
+    if n mod 2 <> 0 then
```
fingolfin, line 47: Below you use `IsOddInt`... Now in GAP probably your code here is faster; but using `IsOddInt` IMHO is easier to grok?
```suggestion
    if IsOddInt(n) then
```

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
+        fi;
+        j := 1;
+        r := s^0;
+        while j <= (n-3)/2  do
+            r := r * s;
+            if not(isone(ri)((t *(t^r))^2)) then
+                return false;
+            fi;
+            j := j + 1;
+        od;
```
fingolfin, line 60: Optimization suggestion (avoids recomputing `(n-3)/2`, and the GAP kernel has a special fast case to compute `for i in RANGE` loops; also saves one multiplication `r * s` per loop iteration; you do need to add `tr` to the list of local variables, though)
```suggestion
        tr := t;
        for j in [1..(n-3)/2] do
            tr := tr ^ s; # equal to t^(s^j)
            if not isone(ri)((t * tr)^2) then
                return false;
            fi;
        od;
```

gap/SnAnBB.gi
```diff
@@ -18,100 +18,917 @@
...
+        j := j + 1;
+    od;
+    return true;
+end);
+
+BindGlobal("SatisfiesAnPresentation",
+function(ri, s, t, n)
+    local j, r, ti;
+
+    if not(isone(ri)(s^(n-2))) or not(isone(ri)(t^3)) then
```
fingolfin, line 43: `not` is not a function...
```suggestion
    if not isone(ri)(s^(n-2)) or not isone(ri)(t^3) then
```
Or course the same applies elsewhere, too

gap/SnAnBB.gi
fingolfin: Argh, only after spending half an hour or more on this did I realize that THIS IS NOT YOUR NEW CODE, you only edited the existing stuff. :-(

I'd really prefer that in a separate PR, clearly marked as such, so that I don't waste time combing through code that is meant to be replaced. Also, I really question the usefulness of reformatting this code; it breaks `git blame` which otherwise is an important debugging tool.

I'll leave my comment in, even though I don't expect anybody to apply them to this old code.

gap/SnAnUnknownDegree.gd
```diff
@@ -0,0 +1,12 @@
+DeclareGlobalName("AllPrimesUpTo");
```
fingolfin, line 1: Don't use `DeclareGlobalName`, it doesn't work in GAP <= 4.11.

Might be best to just drop this file?

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+        fi;
+    od;
+    if not success then
+        return fail;
+    fi;
+    # case distinction on F as in the table of Algorithm 4.20
+    # via a decision tree
+    if F4[1] then
+        if F4[2] then
+            # We are in the 1. Case, since the 2. Case is handled during the
```
fingolfin, line 336: Change `Case` -> `case` everywhere. Also the `1. case` reads really German. I'd write `first case`, or (more likely) `case 1`.

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+BindGlobal("BolsteringElements",
+function(ri, c, eps, N)
+    local result, R, S, nrPrebolsteringElms, i, r, cr, cr2;
+    result := [];
+    R := Int(Ceil(7 / 4 * Log2(Float(eps ^ -1))));
+    S := 7 * N * R;
+    nrPrebolsteringElms := 0;
+    i := 0;
+    # find pre-bolstering elements
+    while i <= S and nrPrebolsteringElms <= R do
```
fingolfin, line 172: Isn't `nrPrebolsteringElms = Length(result)` a loop invariant? Then you could get rid of `nrPrebolsteringElms` and turn this into a for loop:
```suggestion
    for i in [0..S] do
```
and move the `nrPrebolsteringElms <= R` check into the loops...

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+                and arecommuting(ri)(cr2, c)
+        then
+            nrPrebolsteringElms := nrPrebolsteringElms + 1;
+            if isone(ri)((cr ^ (c * r)
+                      * cr ^ (cr2 * c)) ^ 3)
+            then
+                Add(result, c ^ 2 * r);
+            else
+                Add(result, cr);
+            fi;
```
fingolfin, line 189:
```suggestion
            fi;
            if Length(result) > R then break; fi;
```

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+    return [g, 2 * mDash + 2 * m + 3];
+end);
+
+# ri : recog info record with group G
+# c : element of G,
+#     should be a 3-cycle
+# eps : real number, the error bound
+# N : integer, upper bound for the degree of G
+#
+# Returns fail or a list consisting of:
```
fingolfin, line 554:
```suggestion
# Returns fail or a list [g, k] consisting of:
```

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+# - g: element of G,
+#      should be a k-cycle matching c
+# - k: integer,
+#      should be length of cycle g.
+#
+# Note that the order of the returned list is reversed wrt the paper to be
```
fingolfin, line 560:
```suggestion
# Note that the order of the returned list is reversed with respect to the paper to be
```

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,713 @@
...
+            fi;
+        od;
+        return true;
+    end;
+    cg := c ^ g;
+    cg2 := cg ^ g;
+    cg3 := cg2 ^ g;
+    cg4 := cg3 ^ g;
+    H1 := [ c ^ 2, c ^ cg, ~[2] ^ cg3, ~[3] ^ cg3, ~[4] ^ cg4 ];
+    # Test whether an elm of the set X commutes with at least
```
fingolfin, line 238:
```suggestion
    # Test whether an element of the set X commutes with at least
```

fingolfin (commented):
Thank you once again for working on this!

I left some nitpicking comments, as my brain was/is too tired for more. I'll try to look with a more mathematical view in the coming days.

gap/SnAnBB.gd
```diff
@@ -0,0 +1,9 @@
+DeclareGlobalName("SatisfiesSnPresentation");
```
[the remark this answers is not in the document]
ssiccha: I only thought it were good practice to have the function names declared in a .gd file. I always do that out of habit. We don't need either of the two .gd files we added so I'll simply remove them.

gap/SnAnBB.gi
ssiccha: Argh, sorry. Will do next time. I called the commit "Uncomment most functions in gap/SnAnBB.gi" but then ofc the standard "diff view" on github doesn't show the commit messages. :/

gap/SnAnUnknownDegree.gd
ssiccha: Done

gap/SnAnUnknownDegree.gi
ssiccha: done

gap/SnAnUnknownDegree.gi
ssiccha: done

gap/SnAnUnknownDegree.gi
ssiccha: done

init.g
```diff
@@ -21,6 +21,9 @@ ReadPackage("recog","gap/base/recognition.gd");
 # The following contain generic declarations for different types of groups:
 ReadPackage("recog","gap/base/projective.gd");
 
+ReadPackage("recog","gap/SnAnUnknownDegree.gd");
+ReadPackage("recog","gap/SnAnBB.gd");
+
```
fingolfin, line 26:
```suggestion
```
These files don't exist anymore

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,710 @@
...
+
+    # Constants
+    # TODO: better iteration over primes
+    M := 1;
+    allPrimes := AllPrimesUpTo(N);
+    for i in [2 .. Length(allPrimes)] do
+        p := allPrimes[i];
+        M := M * p ^ LogInt(N, p);
+    od;
+    # FIXME: Probably B can be chosen smaller
```
fingolfin, line 68: Where does this opinion come from / how is it justified?

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,710 @@
...
+        if nrTriedConjugates >= C or nrThreeCycleCandidates >= T then
+            r := RandomElm(ri, "simplesocle", true)!.el;
+            a := 0;
+            tPower := r ^ M;
+            # Invariant: tPower = (r ^ M) ^ (2 ^ a)
+            repeat
+                a := a + 1;
+                tPowerOld := tPower;
+                tPower := tPower ^ 2;
+            until a = logInt2N or isone(ri)(tPower);
```
fingolfin, line 110: Minor tweak, since for loops are optimized in GAP:
```suggestion
            tPower := r ^ M;
            # Invariant: tPower = (r ^ M) ^ (2 ^ a)
            for a in [1..logInt2N]
                tPowerOld := tPower;
                tPower := tPower ^ 2;
                if isone(ri)(tPower) then break; fi;
            od;
```

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,710 @@
...
+    t := fail;
+
+    # Constants
+    # TODO: better iteration over primes
+    M := 1;
+    allPrimes := AllPrimesUpTo(N);
+    for i in [2 .. Length(allPrimes)] do
+        p := allPrimes[i];
+        M := M * p ^ LogInt(N, p);
+    od;
```
fingolfin, line 67: The value `M` you computer here only depends on the positive integer `N`, right? And you recompute it here again and again. This suggests to me that perhaps using `MemoizePosIntFunction` here might be beneficial, esp. since `ThreeCycleCandidatesIterator` is likely to be called repeatedly. At the same time, N likely isn't going to be very big, is it? (like, often  a few ten thousand perhaps?)

Actually, that reminds me: we should also think about how the heuristic for coming up with an initial value for N could like? For permutation groups, of course it's easy (`NrMovedPoints(G)`). What does one do for matrix groups over a finite field? A naive approach is this: for a subgroup of GL(d,q),  we must have that Factorial(N) divides the size of that GL(d,q), which we know, so we can compute the largest N with this property. But I assume one can do much better (e.g. using representation theory,  knowledge about element orders, ...).

gap/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,710 @@
...
+            r := RandomElm(ri, "simplesocle", true)!.el;
+            a := 0;
+            tPower := r ^ M;
+            # Invariant: tPower = (r ^ M) ^ (2 ^ a)
+            repeat
+                a := a + 1;
+                tPowerOld := tPower;
+                tPower := tPower ^ 2;
+            until a = logInt2N or isone(ri)(tPower);
+            if a = logInt2N then
```
fingolfin, line 111: But that means the last loop iteration is pointless: even if tPower ends up being the identity in it, we'll return here anyway. Is that the intention? Perhaps this test would be better:
```suggestion
            if not isone(ri)(tPower) then
```

gap/perm/giant.gi
```diff
@@ -548,9 +546,7 @@ end;
 
 
 ##  SLP for pi from (1,2,3), sigma
-
 RECOG.SLPforAn :=  function( n, pi )
-
```
fingolfin, line 553: Might want to drop the diffs in giant.gi?

gap/SnAnUnknownDegree.gi
ssiccha: @FriedrichRober  mentioned sometime that @aniemeyer may know a trick to choose a smaller B. We still have to set up a proper mail to ask her that though. :)

gap/perm/giant.gi
ssiccha: Yepp

ssiccha:
Thanks for the suggestions!

ssiccha:
In #28 I added an example or test (I don't remember which one it was) which we should be able to recognize efficiently with the unknown degree method.

I'm also not so sure anymore whether we should put the "unknown degree recognition" into the projective method databases or whether we should call it from inside recognition of classical or almost simple groups. According to #17 the "known degree recognition" was hooked up previously in `almostsimple.gi:933`.

For permutation groups it might make sense to put it into the method database.

fingolfin:
This needs to be resolved resp. master be merged into it, as there are conflicts in `gap/perm.gi`, which also prevent the CI from running

gap/base/recognition.gi
```diff
@@ -161,7 +161,20 @@ InstallGlobalFunction( EmptyRecognitionInfoRecord,
...
+      if IsObjWithMemory(x) then
+        a := StripMemory(x);
+      else
+        a := x;
+      fi;
+      if IsObjWithMemory(y) then
+        b := StripMemory(y);
+      else
+        b := y;
+      fi;
```
ssiccha, line 175: @FriedrichRober You can just do this, since `StripMemory` does nothing when its input doesn't have memory.
```suggestion
      a := StripMemory(x);
      b := StripMemory(y);
```

FriedrichRober: You are absolutely right :)

ssiccha:
Rebased onto master.

ssiccha:
Rebased on master

ssiccha:
The next step is to get the tests in shape. Then I think this should be ready for review.

ssiccha:
I double-checked that all review comments were addressed. Next stop: tests.

ssiccha:
This PR is now ready for review. It implements the algorithm by Jambor et al. It does _not_ integrate that algorithm into recog yet though. For that @FriedrichRober and I are working on the follow up PR #265, since we have to implement some extensions first.

ssiccha:
Argh, I've added some more tests and discovered a bug. I'll push shortly.

ssiccha:
I added a few tests to increase the code coverage. Turns out, there is a bug in the code handling alternating groups.

FriedrichRober:
@sergio, we forgot to set `foundOddPermutation := false;` at the beginning of the function `RECOG.ConstructSnAnIsomorphism`.

ssiccha:
@FriedrichRober fixed.

ssiccha:
We should have a look again at whether code coverage is good. I think it shouldn't be too hard to get a good coverage, now that we're at it anyways.

gap/SnAnBB.gi
```diff
@@ -18,100 +18,906 @@
 ##  The code is based upon the algorithm presented in the paper [BLGN+03].
 ##
 ##
-# DeclareInfoClass( "InfoRecSnAn" );
-# SetInfoLevel( InfoRecSnAn, 0 );
-#
-# # The following are used for comparisons:
```
fingolfin, line 24: In this file, did you only uncomment lines, or did you also perform changes afterwards?

Related to this: If you think you are basically done with this PR, perhaps you could now clean up its history.  IMHO it is not very useful right now (actually, to me, it seems completely useless, but you might disagree; I am happy to discuss this :-).

What I'd do is probably to is this:
1. make sure latest `master` and latest version of this branch are present
2. create a temporary branch which just undoes the commenting out of this file:
    ```
    git checkout -b HACK master 
    perl -pi -e 's/^# //' gap/SnAnBB.gi # undo the commenting out
    git commit -m "Undo commenting out of unused code in gap/SnAnBB.gi" gap/SnAnBB.gi
    ```
    Actually, having done this, you probably want to do a variant of this, were you leave `RecogniseSnAn` and `NiceGeneratorsSnAn` commented out; and then in another commit (before or after this one), move `NiceGeneratorsSnAn` to the end of the file, to match what you did in this PR.
3. switch to `frss/AnSnRecogUnknownDegree`, make sure it is a superset of master:
    ```
    git checkout frss/AnSnRecogUnknownDegree
    git merge master
    ```
4. now discard the useless history and start over, based on the commit we just made in the HACK branch:
    ```
    git reset HACK
    git add doc gap tst read.g
    git commit
    ```

To make sure you both get credit, add `Co-authored-by: NAME <EMAIL>` to the end of the commit message.

With that, it becomes easy to see what you changed in `gap/SnAnBB.gi`. I just followed the steps above and it was quite helpful, and would also help for reviewing this PR.

Now, of course the above can be further refined; e.g. you might want to split the commits further. But this is up to you.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+        # three cycle candidates for the current involution t.
+        # If this is the case, we need to construct the next involution
+        if nrTriedConjugates >= C or nrThreeCycleCandidates >= T then
+            r := RandomElm(ri, "SnAnUnknownDegree", true)!.el;
+            tPower := r ^ M;
+            # Invariant: tPower = (r ^ M) ^ (2 ^ a)
+            # We make a small improvement to the version described in
+            # <Cite Key="JLNP13"/>. The order of r ^ M is a 2-power.
+            # It can be at most 2 ^ logInt2N. Thus, if we find an r such that
+            # (r ^ M) ^ (2 & logInt2N) is non-trivial, then we can return
```
fingolfin, line 137: What do you mean by `2 & logInt2N`?

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+##
+##  This file provides code for recognising whether a permutation group
+##  is isomorphic to an alternating or symmetric group. It implements
+##  <Cite Key="JLNP13"/>.
+##
+#############################################################################
+#
+# FIXME: Move me into GAP
+# Helper function to compute all primes up to a given integer via a prime sieve
+RECOG.AllPrimesUpTo := function(n)
```
fingolfin, line 23: I am actually not so sure that it is best to move this to GAP? It would then open the question whether GAP should cache these primes or not (which you want), and you'd be tied to that GAP version, and any fixes or improvements would be out of our hands.

I would suggest a slightly different interface anyway:

```gap
RECOG.PrimesCache := ShallowCopy(Primes); # all primes < 1000
RECOG.PrimesCacheUpperBound := 1000;
RECOG.CachePrimesUpTo := function(n)
    local p, ...
    if RECOG.PrimesCacheUpperBound >= n then return; fi;
    # Now run a sieve...
    ...
    RECOG.PrimesCacheUpperBound := n;
end;
```
Instead of storing `RECOG.PrimesCacheUpperBound`, one could also simply replace `n` by `2*n`, and then check `if Last(RECOG.PrimesCache) >= n then return; fi;`: by Bertrand's postulate, there is always a prime between n and 2*. T hen we compute more primes than asked for, but  ...
1. that's not too much of an issue for the numbers we are interested in,
2. you could compensate for this by implementing a common optimization to the sieve, namely skipping all even numbers from the start (halving the require memory for `sieve` again)
3. it actually can be useful if there is bad code which calls this function with N, N+1, N+2, ...: i.e., we kinda implemented "exponential backoff"
4. it can be improved by using refinements of Bertrand's postulate; e.g. it is also known that there is a prime between 3n and 4n for any natural number n; so we can do `n:=4*QuoInt(n+2,3);` which still gives us exponential backoff, just with a factor of 4/3 instead of 2.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+# Returns a record of constants used in ThreeCyclesCanditatesIterator.
+RECOG.ThreeCycleCandidatesConstants := function(eps, N)
+    local M, allPrimes, i, p;
+    # Constants
+    M := 1;
+    allPrimes := RECOG.AllPrimesUpTo(N);
+    for i in [2 .. Length(allPrimes)] do
+        p := allPrimes[i];
+        M := M * p ^ LogInt(N, p);
+    od;
```
fingolfin, line 59: With my suggestions above, this would become:
```suggestion
    RECOG.CachePrimesUpTo(N);
    for p in RECOG.PrimesCache do
        if p > N then break; fi;
        M := M * p ^ LogInt(N, p);
    od;
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+##  <Cite Key="JLNP13"/>.
+##
+#############################################################################
+#
+# FIXME: Move me into GAP
+# Helper function to compute all primes up to a given integer via a prime sieve
+RECOG.AllPrimesUpTo := function(n)
+    local i, j, sieve, result;
+    if n <= 1000 then
+      return Primes{[1..PositionSorted(Primes, n+1)-1]};
```
fingolfin, line 26: Note that this creates a new list object each time it is called, for no good reason (of course this is just a tiny thing compared to the overall effort the algorithms here expend...)

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+    return rec(type := "Sn",
+               isoData := [[b, h], xis, n],
+               slpToStdGens := SLPOfElms(stdGensSnWithMemory));
+end;
+
+# This method is an implementation of <Cite Key="JLNP13"/>. It is the main
+# function of SnAnUnknownDegree.
+# Note that it currently only works for 11 <= <A>n</A>. To make it work with
+# smaller <A>n</A>, we need to include fixes from Jonathan Conder's B.Sc.
+# Thesis "Algorithms for Permutation Groups".
```
fingolfin, line 747: So... are these fixes included? If so: where are they? I searched for comments referencing "Conder" but saw none... Or this more a "TODO" comment?

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -0,0 +1,895 @@
...
+        d := DimensionOfMatrixGroup(G);
+        # Let N be the largest integer such that A_N has a representation of
+        # dimension d.
+        if ri!.projective then
+            # If n >= 9, then the smallest irreducible projective An-module has
+            # dimension n-2, see [KL90], Proposition 5.3.7.
+            # Assume N >= 9 and use the comment above to compute N. If we
+            # arrive at a value < 9 for N, then we must have been in the case N
+            # < 9.
+            # TODO: do we want to use the table for the other cases?
```
fingolfin, line 824:
```suggestion
            # TODO: do we want to use the table from loc.cit. for the other cases?
```

gap/generic/SnAnUnknownDegree.gi
fingolfin: Aha, I just saw that these are implemented in PR #265

FriedrichRober: Yes, this should be changed to a "TODO" comment.

ssiccha:
Somewhere we have an example where An or Sn act on a non-deleted permutation module. That would be a good test case.

gap/generic/SnAnUnknownDegree.gi
ssiccha: That's a typo and should be `2 ^` ^^

gap/generic/SnAnUnknownDegree.gi
ssiccha: Ok, @FriedrichRober is rewriting that ATM. We would move this to `gap/tools.gi` and rename that file to `gap/utils.g`, or do you have a better idea?

gap/generic/SnAnUnknownDegree.gi
ssiccha: done

ssiccha:
We have reduced our PR to a few commits now. Before merging this we would like to do two more things though:
- Split up the last huge commit into several. We have always made extensive use of our history when debugging errors.
- Add more tests, especially for matrix groups

ssiccha:
We should also check how we compare to the `RECOG.RecogniseFDPM` method in gap/projective/AnSnOnFDPM.gi
Maybe we can remove that entirely.

ssiccha:
We have a version of this branch where the `SnAnUnknownDegree` commits were not squashed into one commit here:
https://github.com/ssiccha/recog/tree/frss/AnSnRecogUnknownDegree-backup

ssiccha:
@fingolfin We put a backup of the squashed version of our branch [into this branch on my fork](https://github.com/ssiccha/recog/tree/frss/AnSnRecogUnknownDegree-squashed).

We now pushed a version of this branch which differs from the last only by having several smaller commits instead of the huge `Add algorithm SnAnUnknownDegree` commit. We cleaned up the history and I hope it is understandable now. :)

We're checking now whether there any TODOs left before this can be reviewed.

FriedrichRober:
We also want to take a look if we can reduce the amount of iterations in `BolsteringElements`, since this seems to be a bottleneck, if we have as input an element `c` that is not a three-cycle.

ssiccha:
Actually, I think, except for the tests, all of mine and @FriedrichRober's points from the previous two posts can be addressed in further PRs.

At the moment, testing `Omega(-1,4,3)` and `Omega(0,5,3)` make line 159 of `tst/working/quick/GenericSnAnUnknownDegree.tst` slow. It takes 70 seconds on my machine. I guess that may also be due to `BolsteringElements`.

With respect to the tests: coverage for `SnAnUnknownDegree.gi` is 97%. More or less the only things that were missing were having more matrix groups, more projective groups, and more groups not being isomorphic to Sn or An. I just added tests for these.

So, IMO this can be reviewed as it is now. I'd like to have this merged in before we merge in for example #263. @fingolfin could you have another look?

fingolfin (approved):

PR #180, opened 2020-08-14
unify testfile names

ssiccha:


wilfwilson (commented):
This PR renames:
* `tst/MatTensor.g → tst/mattensor.g`
* `tst/TestClassicalNatural.g → tst/testclassicalnatural.g`
* `tst/OldTestAll.g → tst/testtensor.g`

And adjusts the new `tst/testtensor.g` file to point to the right file. All fine!

However, by looking at the diff, it seems that only two of the files are renamed with `git mv`, whereas `OldTestAll.g` appears to be deleted, with the `tst/testtensor.g` file being newly created (and git not knowing that they were the same).

Is this the case? If not, then everything's fine!

But if not... Obviously the end result would be the same, but for the purposes of looking through the git history, it might be better to have all three files renamed with `git mv`.

To do this, you could have two commits, the first one solely renaming `tst/OldTestAll.g → tst/testtensor.g`, and the second commit renaming the other two files, and editing `tst/testtensor.g` to point to `tst/mattensor.g` rather than `tst/MatTensor.g`.

wilfwilson (approved):
Actually, ignore that, it really doesn't matter for such a small simple file! It's surely not worth the effort to do it 'properly'.

ChrisJefferson:
git mv doesn't actually "move", it is just shorthand for make a new file and delete the old one, then seperately git tries to guess when you have moved something (and sometimes gets it wrong)

wilfwilson:
Thanks for replying @ChrisJefferson - that's really good to know.

And looking back, it's a bit embarrassing how nitpicky I was being 😬

PR #161, opened 2019-07-18
Misc cleanup

fingolfin:

PR #181, opened 2020-08-14
Miscellaneous changes

ssiccha:
General clean up.

Comments out unused code for natural SL recognition.

Adjusts the unused FindHomMethodsMatrix.NaturalSL to only try the constructive recognition method 10 times. Then returns TemporaryFailure.

Move a broken test.

Fix documentation of RecogniseGeneric.

### DIFF
--- a/gap/matrix/slconstr.gi
+++ b/gap/matrix/slconstr.gi
@@ -2481,15 +2481,21 @@ FindHomMethodsMatrix.NaturalSL := function(ri,G)
     fi;
 
 
+  # Try the recognition method
   n:=ri!.dimension;
   q:=Size(ri!.field);
   p:=Factors(q)[1];
   e:=LogInt(q,p);
   grpmem := GroupWithMemory(G);
-  repeat
-      data:=SLCR.SLDataStructure(grpmem,p,e,n);
-  until data <> fail;
-  genlist:=List(data.gens,x->x[1]);
+  # TODO find a reasonable number of tries
+  for i in [1 .. 10] do
+    data := SLCR.SLDataStructure(grpmem, p, e, n);
+    if data <> fail then break; fi;
+  od;
+  if data = fail then
+    return TemporaryFailure;
+  fi;
+  genlist := List(data.gens, x->x[1]);
 
   # Now clean out the memory everywhere in data:
   for i in [1..Length(data.gens)] do
@@ -2507,6 +2513,7 @@ FindHomMethodsMatrix.NaturalSL := function(ri,G)
   data.t21invtran := StripMemory(data.t21invtran);
   data.sinv := StripMemory(data.sinv);
 
+  # Return
   ri!.data:=data;
   ri!.dimension:=n;
   Setslpforelement(ri,SLPforElementFuncsMatrix.SLConstructive);
--- a/tst/working/quick/ProjSubfield3.tst
+++ b/tst/working/quick/ProjSubfield3.tst
@@ -7,4 +7,7 @@ gap> g := GroupWithGenerators(gens);;
 
 # Remark: The *4 comes from the fact that by blowing up we lose a factor
 #         of 4 in scalars, they are now in the projective group.
+gap> level := InfoLevel(InfoOrb);;
+gap> SetInfoLevel(InfoOrb, 0);;
 gap> ri := RECOG.TestGroup(g,true,Size(PGL(3,3^2)));;
+gap> SetInfoLevel(InfoOrb, level);;


tst/broken/quick/ProjSubfield3.tst
```diff
@@ -4,7 +4,4 @@ gap> gens := List(GeneratorsOfGroup(h),y->y^x);;
 gap> gens[1] := gens[1] * Z(3^4);;
 gap> gens[2] := gens[2] * Z(3^4)^17;;
 gap> g := GroupWithGenerators(gens);;
-
-# Remark: The *4 comes from the fact that by blowing up we lose a factor
-#         of 4 in scalars, they are now in the projective group.
```
wilfwilson, line 9: (Looking at commit `tst: move broken test ProjSubfield3.tst`):

How has this test broken? In the `master` branch, this test runs fine.

gap/base/recognition.gd
```diff
@@ -564,7 +564,7 @@ DeclareSynonym("RecognizeGroup", RecogniseGroup);
 ##     which is created for the group <A>H</A>. Especially the component
 ##     <C>hints</C> can contain a list of additional find homomorphism
 ##     methods (described by records as in Section <Ref Sect="whataremethods"/>)
-##     which is prepended to the method database in <A>methoddb</A> before
+##     which is sorted into the method database in <A>methoddb</A> before
```
wilfwilson, line 567: What does 'sorted into' mean? It's not clear to me.

wilfwilson (commented):
Two questions, otherwise fine.

tst/broken/quick/ProjSubfield3.tst
ssiccha: It passes very often. But if I run it in a session in a while loop it breaks after maybe a second. It also did break on travis at least once, see e.g. older travis runs for #161.

gap/base/recognition.gd
ssiccha: Yeah I wasn't sure whether that'd be obvious. Apparently not. :) I'll clarify that. Thanks!

wilfwilson (approved):

ssiccha:
This failed in Travis
```

working/quick/Sporadic.tst:59

# Input is:

TestSporadic("ON.2");

# Expected output:

[ "ON.2" ]

# But found:

[ "B" ]

########
```
Maybe another test that only breaks sometimes?

wilfwilson:
Surely it would've been noticed before, if it was failing sporadically. And it doesn't seem very sporadic if it has failed several times with this PR. What changes in your PR could have 'caused' this failure? Is it perhaps the `TemporaryFailure` case that you've added, that's leading this test down a different code path, whereupon there's a failure?

ssiccha:
I did a bid of clean-up in a function where I checked afterwards whether it is even called. And it never is. That's the part where I added the `TemporaryFailure`-bit.

Wrt the travis failure: I can reproduce the `TestSporadic` failure on my machine both with the current master of recog and in a version from July 2019. I'm gonna check with two different versions of gap now.

ssiccha:
So, I ran the test again with gap 4.10 and the current master and it also failed there.

ssiccha:
See #186

fingolfin:
When the Travis CI tests run last (already a month ago :/) they passed. Weird. I've closed&reopened this to re-run them ocne more. I'll also (re)activate a Travis CI daily cron job

gap/matrix/classical.gi
```diff
@@ -2257,15 +2257,8 @@ function( arg )
...
                    isSOContained := "unknown",
                   );
-  merkinfolevel := InfoLevel(InfoMethSel);
-  SetInfoLevel(InfoMethSel,0);
-  ret := CallMethods( ClassicalMethDb, nrrandels, recognise, grp );
-  SetInfoLevel(InfoMethSel,merkinfolevel);
-  # fail: bedeutet, dass entnervt aufgegeben wurde
-  # true: bedeutet, dass eine Methode "erfolgreich" war
-
+  CallMethods( ClassicalMethDb, nrrandels, recognise, grp );
```
fingolfin, line 2260: The commit "matrix: clean up RecogniseClassical" is mislabeled; it drops a change to the info level.  I wouldn't call this a "clean up" (for me, a "clean up" is a change that does not change observable behavior, just restructures the code).

As to the change itself: I am not yet sure whether it's a great idea or a horrible one... the current code certainly is a hack, though, because it (ab?)uses `CallMethods`; but all the "methods" that are installed there still return values like `fail` and `false` and `true` (and I am not even sure with the correct semantics!).

gap/matrix/slconstr.gi
```diff
@@ -2466,56 +2466,52 @@ end;
...
+    local data, e, genlist, grpmem, i, j, n, p, q, r, il;
     if not IsMatrixGroup(G) then
         return NeverApplicable;
     fi;
-
-    il := InfoLevel(InfoMethSel);
-    SetInfoLevel(InfoMethSel,0);
-    r:=RecogniseClassical(G,"linear");
-    SetInfoLevel(InfoMethSel,il);
+    r := RecogniseClassical(G, "linear");
```
fingolfin: Again a change that potential is not a "clean up". Well, it would be if you didn't touch `RecogniseClassical` as that already messes with `InfoMethSel` in the same way...

gap/matrix/slconstr.gi
```diff
@@ -2466,56 +2466,52 @@ end;
...
     if not( (IsBound(r.isSLContained) and r.isSLContained = true) ) then
         return TemporaryFailure;
     fi;
-
-
-  n:=ri!.dimension;
-  q:=Size(ri!.field);
-  p:=Factors(q)[1];
```
fingolfin, line 2487: Are large scaler whitespace and indentation changes really going to help us? They do wreck havoc on `git blame`, though...

gap/base/recognition.gd
```diff
@@ -563,9 +563,12 @@ DeclareSynonym("RecognizeGroup", RecogniseGroup);
 ##     components of which are copied into the new recognition info record
 ##     which is created for the group <A>H</A>. Especially the component
 ##     <C>hints</C> can contain a list of additional find homomorphism
-##     methods (described by records as in Section <Ref Sect="whataremethods"/>)
-##     which is prepended to the method database in <A>methoddb</A> before
-##     the recognition starts. This feature is intended to give hints
+##     methods (described by records as in Section <Ref Sect="whataremethods"/>).
+#      The methods in <C>hints</C> and in <A>methoddb</A> are merged and sorted
+#      into rank-descending order. The result is passed to
+#      <Ref Func="CallMethods"/>. Thus you can force the methods in
```
fingolfin, line 569: This sound strange to me. You "can" force this? Isn't it always true that `hints` methods are called first? At the same time, this is only one of the ways you can use `hints` for, you can also use it to add methods which are not applicable in general, but nevertheless might not have the highest rank. So why call out that use case especially? But if you really think it's useful, I'd still rephrase it...

gap/matrix/slconstr.gi
```diff
@@ -2068,126 +2068,127 @@ end;
...
-   if vectors = fail then
-return fail;
-   fi;
-   data3 := SLCR.SLConstructBasisQ(bbg, data2, vectors[1], vectors[2], d, q);
-   data2 := SLCR.SLFinishConstruction(bbg,data2,data3,d);
-return data2;
-end;
+#
+#SLCR.SLDataStructure := function( bbg, p, e, d )
+#    local i, j, stop, q, r, t, u1, u2, sl3, data1, Q, pQ, Lgen,
```
fingolfin, line 2073: I am not a fan of commenting out (*and* reformatting, at the same time...) this code. Alice and me want to work on it soon (assuming our grant proposal is accepted0

gap/base/methsel.gd
```diff
@@ -19,7 +19,7 @@
 # Our own method selection code:
 
 DeclareInfoClass( "InfoMethSel" );
-SetInfoLevel(InfoMethSel,1);
+SetInfoLevel(InfoMethSel, 0);
```
fingolfin, line 22: The default level is 0 anyway, so you could just drop this line:
```suggestion
```
However, I note that there is only one situation when we show `InfoMethSel ` messages on level 1, namely when `CallMethods` went over all methods and nothing worked. That "normally" shouldn't happen (unless we encounter a case that we didn't implement yet), so I kinda see why they set this to level 1 by default -- recog is still very incomplete and under development and seeing this output is a valuable clue for what's wrong.

So I feel uneasy about this change. Perhaps we should discuss this and rest on the phone tomorrow.

gap/matrix/classical.gi
fingolfin: It's also interesting to see that the return value of `CallMethods` is completely ignored.

fingolfin (changes requested):

gap/matrix/classical.gi
ssiccha: >  I wouldn't call this a "clean up" (for me, a "clean up" is a change that does not change observable behavior, just restructures the code).

Good to know. I also included changing the debugging output into "clean up".

> As to the change itself: I am not yet sure whether it's a great idea or a horrible one...

I'm a bit confused. I only changed the infolevels and got rid of the unused `ret` variable. Surely, only changing the infolevels can't be that bad. Or am I missing something?

I guess when I get to the classical groups I will try to reorganize the code and get rid of that `CallMethods` call.

ssiccha: I mean, all we would possibly do is print _more_ information. Ofc I could add something like `Info(InfoMethsel, 2, "HACK: CallMethods with ClassicalMethDb");`

gap/matrix/slconstr.gi
ssiccha: I was looking through the function and kinda got carried away. But yeah, I should try more not to spend time on changing whitespace only. In my defense, the indentation looked reaally wonky. ^^

gap/matrix/slconstr.gi
ssiccha: Ok, I'll drop the commit.

gap/base/methsel.gd
ssiccha: Makes sense. I'm dropping that commit too.

gap/base/recognition.gd
ssiccha: I'll drop the "can force" sentence. It's not like it adds anything valuable.

> Isn't it always true that hints methods are called first? 

This is why I actually did this commit. At the moment, the hint methods are _not_ always called first, despite what the documentation says. This is the responsible [line of code](https://github.com/gap-packages/recog/blob/678f1317332e152fe5d718231a0ecdb78e4872ab/gap/base/recognition.gi#L420). But I ran git blame a little and I have the feeling that the original "sorting the hints and methoddb into rank-descending order" was done on accident in commit 30717370 with title "Add example of successful recognition". I guess that commit should only have touched  `doc/afterrecog.xml`.

gap/matrix/classical.gi
fingolfin: Exactly, we'd be printing more information, potentially *lots* more information that one can not suppress anymore now, without also suppressing information about the "top level" method selection: because here a leaf recognition method uses `CallMethods` itself, which no other method does.

So I am concerned that some debugging might get considerably harder because the information one is interested in is hidden in a flood of info messages that weren't there before, and which I can't turn off without also turning off the messages I *am* interested in...

ssiccha: My thinking was that since `RecogniseGeneric` calls itself recursively, it calls `CallMethods` on every node of the "recognition tree" anyways.
I guess it makes more sense for me to look at this again when I have some concrete examples or if I'm rewriting that code anyway. I'll drop the respective commit.

ssiccha: I've moved the respective commit into PR #187

gap/base/recognition.gd
ssiccha: I've moved the respective commit into PR #187

gap/matrix/slconstr.gi
```diff
@@ -2482,16 +2482,21 @@ FindHomMethodsMatrix.NaturalSL := function(ri,G)
...
+    data := SLCR.SLDataStructure(grpmem, p, e, n);
+    if data <> fail then break; fi;
+  od;
+  if data <> fail then
+    genlist := List(data.gens, x->x[1]);
+    data.slpnice := SLPOfElms(genlist);
+  fi;
 
   # Now clean out the memory everywhere in data:
   for i in [1..Length(data.gens)] do
```
fingolfin, line 2502: Won't this run into an error if `data = fail`? I think the `data = fail` test needs to be moved before this.

tst/broken/quick/ProjSubfield3.tst
```diff
@@ -1,10 +1,8 @@
+# This test breaks if you run it e.g. 50 times in a row.
```
fingolfin, line 1: Hmm, I tried this, and yes it failed: but the issue was that it unexpectedly said:
```
#I  Giving up, Schreier tree is not shallow.
```

This message is from `orb`, and can be disabled via `SetInfoLevel(InfoOrb, 0);`; so we could just do that before the last line of this test (and then restore the old info level afterwards)

ssiccha: Done. I'm now gonna finish the actions PR.

ssiccha:
Rebased onto master.

I think this PR should now be ready @fingolfin.

gap/matrix/slconstr.gi
ssiccha: Done.

fingolfin (approved):

PR #186, opened 2020-10-19
Miscellaneous clean up

ssiccha:


ssiccha:
See discussion in #181

tst/broken/quick/Sporadic.tst
```diff
@@ -28,6 +28,8 @@ gap> TestSporadic("HN.2"); # FIXME: sometimes gives "B"
 [ "HN.2" ]
 gap> TestSporadic("HS.2"); # FIXME: sometimes "recognized" as McL.2
 [ "HS.2"]
+gap> TestSporadic("ON.2"); # FIXME: sometimes gives "B"
+[ "ON.2" ]
```
fingolfin, line 32: Actually this doesn't fail on Traivs CI right now?

But anyway, I can reproduce an occasional failure here locally. And I am not surprised by it either. So, sure, let's do that for now.

fingolfin (approved):

PR #187, opened 2020-10-22
Fix RecogniseGroup wrt sorting hints

ssiccha:
Either the documentation of `RecogniseGroup` is incorrect, or the function
itself is incorrect. See the discussion in #181.

### DIFF
--- a/gap/base/recognition.gd
+++ b/gap/base/recognition.gd
@@ -563,9 +563,11 @@ DeclareSynonym("RecognizeGroup", RecogniseGroup);
 ##     components of which are copied into the new recognition info record
 ##     which is created for the group <A>H</A>. Especially the component
 ##     <C>hints</C> can contain a list of additional find homomorphism
-##     methods (described by records as in Section <Ref Sect="whataremethods"/>)
-##     which is prepended to the method database in <A>methoddb</A> before
-##     the recognition starts. This feature is intended to give hints
+##     methods (described by records as in Section <Ref Sect="whataremethods"/>).
+#      The methods in <C>hints</C> and in <A>methoddb</A> are merged and sorted
+#      into rank-descending order. The result is passed to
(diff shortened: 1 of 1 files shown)


ssiccha:
I've dropped the "Thus you can force..." sentence.

ssiccha:
@fingolfin What do you think of this?

> At the moment, the hint methods are not always called first, despite what the documentation says. This is the responsible line of code. But I ran git blame a little and I have the feeling that the original "sorting the hints and methoddb into rank-descending order" was done on accident in commit 3071737 with title "Add example of successful recognition". I guess that commit should only have touched doc/afterrecog.xml.

ssiccha:
I just checked which ranks we use to install hints. There are two different situations. Either we choose the rank such that the hint is called before all methods from the respective method database. Or we choose the rank such that the hint is called after the method `TrivialGroup` but before all other methods from the respective database.

If we want to keep it such that `TrivialGroup` can be called first, then we should go with the current form of the code, that is concatenating hints and the database and then sorting. Then the documentation would need to be updated.

fingolfin:
I think the new methods should be prepended and then all sorted in (possibly via `StableSort`, to ensure that in case of a tie in rank, the hint is first; but probably even better would be to have a test which checks that all ranks are different and if they are not, show an error).

The reason is this: if I want all my hints to be sorted first, I can trivially achieve this by using suitable ranks. But if I want my hints to be interspersed with some existing methods, I cannot do this anymore if we just blindly prepend, unless I am copy&pasting existing methods into the table.

And then we might as well drop those numeric hints completely and just require that the methods tables are set up once at the start with everything in the desired order... To be clear I don't think we should do that: while ranks are somewhat arbitrary, at least they give us a semiclean way to insert extra methods in multiple places, i.e., they help with modularity.

fingolfin (approved):

PR #189, opened 2020-10-23
Miscellaneous clean up

ssiccha:


fingolfin:
Now there is a CI failure:
```
working/quick/Sporadic.tst:43
# Input is:
TestSporadic("Suz");
# Expected output:
[ "Suz" ]
# But found:
[ "Fi22.2" ]
```
But I think that's just another case of an instable recognition, so you can deal with it as usual :-)

fingolfin (approved):

PR #191, opened 2020-10-23
Update ThrowAwayFixedPoints

### LINKED ISSUE #190: Large Base recognition fails although it shouldn't
The large base group, or jellyfish group, recognition is not able to cope with a group which has fixed points.

Consider the following two examples. The first one works, the second is a conjugate group with fixed points which doesn't:
```
gap> S23OnSets := Action(SymmetricGroup(23), Combinations([1 .. 23], 2), OnSets);;
gap> jellyGroup2 := WreathProductProductAction(S23OnSets, Group((1,2)));;
gap> n := NrMovedPoints(jellyGroup2);
64009
gap> ri2 := RecogniseGroup(jellyGroup2);
F pts= 64009  0       
<recoginfo LargeBasePrimitive
 F:<recoginfo Imprimitive
    F:<recoginfo Pcgs Size=2>
    K:<recoginfo BalTreeForBlocks
       F:<recoginfo Giant AlmostSimple Size=25852016738884976640000>
       K:<recoginfo Giant AlmostSimple Size=25852016738884976640000>>>
 K:<trivial kernel>
gap> # Now take a conjugate
gap> random := Random(SymmetricGroup(5 * QuoInt(n,2)));;
gap> jellyGroup2Conj := jellyGroup2 ^ random;;
gap> ri2Conj := RecogniseGroup(jellyGroup2Conj);
#I  Have 32065 points.
#I  Have 60697 points.
#I  Have 57149 points in new orbit.
#I  Have 64002 points in new orbit.
#I  Have 58357 points in new orbit.
#I  Have 64003 points in new orbit.

<recoginfo StabilizerChainPerm Size=1336653538935178044929642368586691379200000000>
gap> ri2Conj!.fhmethsel.failedMethods.LargeBasePrimitive;
1
```

ssiccha:
Also lets ThrowAwayFixedPoints create a homomorphism if the input group
is primitive and has fixed points.

Fixes #190.

### DIFF
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -355,17 +355,17 @@ end;
 #! @BeginChunk ThrowAwayFixedPoints
 #! This method defines a homomorphism of a permutation group
 #! <A>G</A> to the action on the moved points of <A>G</A> if
-#! either <A>G</A> has a lot of fixed points, or if <A>G</A> knows that it is
-#! primitive and has fixed points. If <A>G</A> has fixed points but does not
-#! know whether it is primitive, then it returns <K>NotEnoughInformation</K>
-#! so that it may be called again at a later time.
+#! <A>G</A> has any fixed points, and is either known to be primitive or the
+#! ratio of fixed points to moved points exceeds a certain threshold. If <A>G</A>
+#! has fixed points but does not meet either of these criteria, then it returns
+#! <K>NotEnoughInformation</K> so that it may be called again at a later time.
 #! In all other cases, it returns <K>NeverApplicable</K>.
 #! <P/>
 #!
 #! In the current setup, the
 #! homomorphism is defined if the number <M>n</M> of moved
 #! points is at most <M>1/3</M> of the largest moved point of <A>G</A>,
-#! or  <M>n</M> is at most half of the number of points on which
+#! or <M>n</M> is at most half of the number of points on which
 #! <A>G</A> is stored internally by &GAP;.
 #! <P/>
 #!
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -355,24 +355,44 @@ end;
 #! @BeginChunk ThrowAwayFixedPoints
 #! This method defines a homomorphism of a permutation group
 #! <A>G</A> to the action on the moved points of <A>G</A> if
-#! <A>G</A> does not have too many moved points. In the current setup, the
-#! homomorphism is defined if the number <M>k</M> of moved
+#! either <A>G</A> has a lot of fixed points, or if <A>G</A> knows that it is
+#! primitive and has fixed points. If <A>G</A> has fixed points but does not
+#! know whether it is primitive, then it returns <K>NotEnoughInformation</K>
+#! so that it may be called again at a later time.
+#! In all other cases, it returns <K>NeverApplicable</K>.
+#! <P/>
+#!
+#! In the current setup, the
+#! homomorphism is defined if the number <M>n</M> of moved
 #! points is at most <M>1/3</M> of the largest moved point of <A>G</A>,
-#! or  <M>k</M> is at most half of the number of points on which
-#! <A>G</A> is stored internally by &GAP;. The method returns
-#! <K>NeverApplicable</K> if it does not define a homomorphism indicating that it will
-#! never succeed.
+#! or  <M>n</M> is at most half of the number of points on which
+#! <A>G</A> is stored internally by &GAP;.
+#! <P/>
+#!
+#! The fact that this method returns <K>NotEnoughInformation</K> if <A>G</A>
+#! has fixed points but does not know whether it is primitive, is important for
+#! the efficient handling of large-base primitive groups by
+#! <Ref Func="LargeBasePrimitive"/>.
 #! @EndChunk
-FindHomMethodsPerm.ThrowAwayFixedPoints :=
-  function( ri, G )
+FindHomMethodsPerm.ThrowAwayFixedPoints := function( ri, G )
       # Check, whether we can throw away fixed points
-      local gens,hom,l,n,o;
+      local gens,nrStoredPoints,n,largest,isApplicable,o,hom;
 
       gens := GeneratorsOfGroup(G);
-      l := List(gens,StoredPointsPerm);
+      nrStoredPoints := Maximum(List(gens,StoredPointsPerm));
       n := NrMovedPoints(G);
-      if 2*n > Maximum(l) or 3*n > LargestMovedPoint(G) then  # we do nothing
-          return NeverApplicable;
+      largest := LargestMovedPoint(G);
+      # If isApplicable is true, we definitely are applicable.
+      isApplicable := 3*n <= largest or 2*n <= nrStoredPoints
+        or (HasIsPrimitive(G) and IsPrimitive(G) and n < largest);
+      # If isApplicable is false, we might become applicable if G figures out
+      # that it is primitive.
+      if not isApplicable then
+          if not HasIsPrimitive(G) and n < largest then
+              return NotEnoughInformation;
+          else
+              return NeverApplicable;
+          fi;
       fi;
       o := MovedPoints(G);
       hom := ActionHomomorphism(G,o);
--- a/gap/perm.gi
+++ b/gap/perm.gi
@@ -357,7 +357,7 @@ end;
 #! <A>G</A> to the action on the moved points of <A>G</A> if
 #! <A>G</A> has any fixed points, and is either known to be primitive or the
 #! ratio of fixed points to moved points exceeds a certain threshold. If <A>G</A>
-#! has fixed points but does not meet either of these criteria, then it returns
+#! has fixed points but is not primitive, then it returns
 #! <K>NotEnoughInformation</K> so that it may be called again at a later time.
 #! In all other cases, it returns <K>NeverApplicable</K>.
 #! <P/>
@@ -374,6 +374,23 @@ end;
 #! the efficient handling of large-base primitive groups by
 #! <Ref Func="LargeBasePrimitive"/>.
 #! @EndChunk
+#
+#  A technical explanation of why we use a threshold for the ratio of fixed
+#  points to moved points, this is technical so it does not go into the manual:
+#  The reason we do not just always discard fixed points is that it also incurs
+#  a cost, so we only try to do it when it seems worthwhile to do so.
+#
+#  Suppose the group G is not known to be primitive. Then we still
+#  apply this method if one of the following two criterion is met:
+#  - the largest moved point of G is three or more times larger than the number
+#    n of actually moved points
+#  - there is a generator of G whose internal storage reserves space for a
+#    number of moved points which is two or more times larger than n. Note that
+#    even for a simple transposition (1,2), for technical reasons it can happen
+#    that GAP internally stores it with the same size as a permutation moving a
+#    million points; this is wasteful, and the second criterion tries to deal
+#    with this.
+
 FindHomMethodsPerm.ThrowAwayFixedPoints := function( ri, G )
       # Check, whether we can throw away fixed points
       local gens,nrStoredPoints,n,largest,isApplicable,o,hom;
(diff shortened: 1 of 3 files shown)


ssiccha:
So, on the upside this fixes #190. On the downside recognition of a large base primitive groups with fixed points takes 20 seconds compared to 2 seconds for an isomorphic group but without fixed points.

Here is an example run:
```
gap> S23OnSets := Action(SymmetricGroup(23), Combinations([1 .. 23], 2), OnSets);;
gap> jellyGroup2 := WreathProductProductAction(S23OnSets, Group((1,2)));;
gap> n := NrMovedPoints(jellyGroup2);
64009
gap> ri2 := RecogniseGroup(jellyGroup2); time;
F pts= 64009  0        
<recoginfo LargeBasePrimitive
 F:<recoginfo Imprimitive
    F:<recoginfo Pcgs Size=2>
    K:<recoginfo BalTreeForBlocks
       F:<recoginfo Giant AlmostSimple Size=25852016738884976640000>
       K:<recoginfo ThrowAwayFixedPoints
          F:<recoginfo Giant AlmostSimple Size=25852016738884976640000>
          K:<trivial kernel>>>
 K:<trivial kernel>
1891
gap> random := Random(SymmetricGroup(5 * QuoInt(n,2)));;
gap> jellyGroup2Conj := jellyGroup2 ^ random;;
gap> ri2Conj := RecogniseGroup(jellyGroup2Conj); time;
F pts=160018  0         
<recoginfo ThrowAwayFixedPoints
 F:<recoginfo LargeBasePrimitive
    F:<recoginfo Imprimitive
       F:<recoginfo Pcgs Size=2>
       K:<recoginfo BalTreeForBlocks
          F:<recoginfo Giant AlmostSimple Size=25852016738884976640000>
          K:<recoginfo ThrowAwayFixedPoints
             F:<recoginfo Giant AlmostSimple Size=25852016738884976640000>
             K:<trivial kernel>>>
    K:<trivial kernel>
 K:<trivial kernel>
22211
```

I'm looking into a profile now and it seems like two things make this so slow
- the `IsPrimitive` check I added is super expensive because computing the blocks takes ages
- the multiplication of group elements with memory becomes very expensive

ssiccha:
Hm, apparently the `IsPrimitive` call in the second level of the tree, that is directly after applying `ThrowAwayFixedPoints` for the first time, takes super long.

gap/perm.gi
```diff
@@ -361,12 +363,15 @@ end;
...
-      local gens,hom,l,n,o;
+      local gens,hom,l,n,largest,o;
 
       gens := GeneratorsOfGroup(G);
       l := List(gens,StoredPointsPerm);
       n := NrMovedPoints(G);
-      if 2*n > Maximum(l) or 3*n > LargestMovedPoint(G) then  # we do nothing
+      largest := LargestMovedPoint(G);
+      if (2*n > Maximum(l) or 3*n > largest)
+              and not (IsPrimitive(G) and largest > n)
```
fingolfin, line 373: It's just too expensive to check `IsPrimitive`.

Instead we should either...
1. enhance the  large base group to also check for fixed points and throw them away (either with its own code or by forcing a `FindHomMethodsPerm.ThrowAwayFixedPoints` step with this limit removed.

2. improve the large base group case to not require that `MovedPoints(G) = [1..LargestMovedPoints(G)]`

fingolfin (changes requested):

gap/perm.gi
fingolfin: @ssiccha any plans to revisit this?

ssiccha:
@fingolfin I just pushed a version that I think does the trick. The tests I added take roughly 17 seconds on my machine. Do you think that is too much for the `quick` testsuite?

ssiccha:
Hmm, does this brake something in `tst/working/quick/ProjLowIndex.tst`?

ssiccha:
The test-suite ran into a failure in `ProjLowIndex.tst`. I was able to reproduce that on my machine by executing `tst/testquick.g`. However, I also ran `ProjLowIndex.tst` a thousand times both with the current master branch and with this PR just to make sure, that this PR does not by some weird quirk introduce a regression. With this PR the test fails three out of a thousand times. In master it fails four out of a thousand times.

So by a big coincidence, when arriving at `ProjLowIndex.tst` in the test suite this PR exactly hit a random state, such that the group in `ProjLowIndex.tst` is recognized incorrectly.

I have thus changed the test in `PermLargeBasePrimitive.tst` to use a smaller group. Which has the benefit that a) the test runs quicker and b) `ProjLowIndex.tst` also passes again.

gap/perm/largebase.gi
```diff
@@ -507,6 +507,14 @@ end;
 #! <M>r \cdot k > 1</M>
 #! and <M>2 \cdot r \cdot k^2 \le n</M>.
 #!
+#! A large primitive group <M>H</M>of the above type which does have fixed
+#! points is handled as follows.  After a call of <C>LargeBasePrimitive</C>,
+#! the group <M>H</M> knows that it is primitive. Thus, after
+#! LargeBasePrimitive returns <K>NeverApplicable</K>, the function
+#! <Ref Func="ThrowAwayFixedPoints"/> will prune away the fixed points and set
+#! up a homomorphism. <C>LargeBasePrimitive</C> is then applicable to the image
+#! of that homomorphism.
```
fingolfin, line 516: This phrasing really confused me. First yo talk about what happens if `LargeBasePrimitive` succeeds; and then you switch to what happens if it is not applicable; in that case it might be that `IsPrimitive` is false; so why then does it imply that `ThrowAwayFixedPoints` will prune away fixed points ?!?! DId you perhaps mean to refer to the return value "TemporaryFailure" ?

Anyway, it all got a bit clearer when I looked at the code and saw that `FindHomMethodsPerm.LargeBasePrimitive` simply *always* invokes `IsPrimitive` on any permutation group fed to it; OK, fair enough, the comment got a bit clearer. But I feel the comment should be made clear on its own.

I'd try to suggest something, but I am not exactly sure what you wanted to express here, so it's a bit difficult. But perhaps something simple: like this:

```suggestion
#! A large primitive group <M>H</M>of the above type which does have fixed
#! points is handled as follows:  After the first call to <C>LargeBasePrimitive</C>,
#! the group <M>H</M> knows whether it is primitive or not. Thus, if
#! LargeBasePrimitive returns <K>TemporaryFailure</K> for a primitive group, the function
#! <Ref Func="ThrowAwayFixedPoints"/> will use the new information about <M>H</M>
#! being primitive, and then is guaranteed to prune away the fixed points and set
#! up a reduction homomorphism. <C>LargeBasePrimitive</C> is then applicable to the image
#! of that homomorphism.
```

tst/working/quick/PermLargeBasePrimitive.tst
```diff
@@ -1,13 +1,29 @@
...
-gap> RECOG.TestGroup(jellyGroupConj,
+gap> ri := RECOG.TestGroup(jellyGroupConj,
 >   false,
 >   Size(jellyGroup),
 >   rec(tryNonGroupElements := true));;
+gap> ri!.fhmethsel.successMethod;
+"LargeBasePrimitive"
+
+# Test that groups with fixed points are handled together by
+# ThrowAwayFixedPoints and LargeBasePrimitive.
```
fingolfin, line 19: Maybe reference this PR and/or issue #190 in the comment, to make it easy to get further details if required?

fingolfin (approved):
Seems sensible. Left some minor comments

gap/perm/largebase.gi
ssiccha: Ok, I'll have a look before we merge this in.

tst/working/quick/PermLargeBasePrimitive.tst
ssiccha: Done.

ssiccha:
@fingolfin I have adjusted and integrated your suggestion into the documentation of `LargeBasePrimitive`. I've also added a short explanation to the documentation of `ThrowAwayFixedPoints`. Could you have a look again to see whether it is now clear how large base primitive groups with fixed points are handled?

gap/perm.gi
```diff
@@ -347,24 +347,44 @@ end;
 #! @BeginChunk ThrowAwayFixedPoints
 #! This method defines a homomorphism of a permutation group
 #! <A>G</A> to the action on the moved points of <A>G</A> if
-#! <A>G</A> does not have too many moved points. In the current setup, the
-#! homomorphism is defined if the number <M>k</M> of moved
+#! either <A>G</A> has a lot of fixed points, or if <A>G</A> knows that it is
+#! primitive and has fixed points. If <A>G</A> has fixed points but does not
+#! know whether it is primitive, then it returns <K>NotEnoughInformation</K>
+#! so that it may be called again at a later time.
```
fingolfin, line 353: This section confused me a bit. How about something more like this?
```suggestion
#! <A>G</A> has any fixed points, and is either known to be primitive or the
#! ratio of fixed points to moved points exceeds a certain threshold. If <A>G</A>
#! has fixed points but does not meet either of these criteria, then it returns
#! <K>NotEnoughInformation</K> so that it may be called again at a later time.
```

gap/perm.gi
```diff
@@ -347,24 +347,44 @@ end;
...
+#! <P/>
+#!
+#! In the current setup, the
+#! homomorphism is defined if the number <M>n</M> of moved
 #! points is at most <M>1/3</M> of the largest moved point of <A>G</A>,
-#! or  <M>k</M> is at most half of the number of points on which
-#! <A>G</A> is stored internally by &GAP;. The method returns
-#! <K>NeverApplicable</K> if it does not define a homomorphism indicating that it will
-#! never succeed.
+#! or  <M>n</M> is at most half of the number of points on which
```
fingolfin, line 360:
```suggestion
#! or <M>n</M> is at most half of the number of points on which
```

gap/perm.gi
```diff
@@ -347,24 +347,44 @@ end;
...
+#! <A>G</A> is stored internally by &GAP;.
```
fingolfin, line 361: Perhaps this should be explained a bit more, but then it *is* a bit technical... Perhaps what I wrote above could be extended... something along these lines:

> The reason we do not just always discard fixed points is that it also incurs a cost, so we only try to do it when it seems worthwhile to do so.
>
> To explain what was meant above by the ration of fixed points to moved points, suppose the group G is
> not known to be primitive. Then we still apply this method if one of the following two criterion is met:
> 1. the largest moved point of G is three or more times larger than the number n of actually moved points 
> 2. there is a generator of G whose internal storage reserves space for a number of moved points which is two or more times larger than n (note that even for a simple transposition (1,2), for technical reasons it can happen that GAP internally stores it with the same size as a permutation moving a million points; this is wasteful, and the second criterion tries to deal with this)

gap/perm/largebase.gi
```diff
@@ -526,6 +538,7 @@ FindHomMethodsPerm.LargeBasePrimitive :=
     fi;
     RECOG.SetPseudoRandomStamp(grp,"Jellyfish");
     res := RECOG.AllJellyfish(grp);
+    RECOG.SetPseudoRandomStamp(grp,"PseudoRandom");
```
fingolfin, line 541: Why not this? (Honest question, I don't know)
```suggestion
    RECOG.SetPseudoRandomStamp(grp,"LargeBasePrimitive");
```

fingolfin (approved):
Fine in principle, thank you! I do have some suggestions for the text, though.

gap/perm/largebase.gi
ssiccha: `LargeBasePrimitive` is more or less a wrapper for `AllJellyfish`. After the call to `AllJellyfish` the actual computations are done and now we're just managing stuff. Thus here I "reset" the stamp.

Yeah, I really do not like how we handle this whole RandomStamp business currently, see #261 for an idea how to improve this.

gap/perm.gi
ssiccha: I agree that that's a bit technical. So I'll put it into a source comment but not into the manual.

ssiccha (commented):
I've added a commit with your suggestions and one with some minor changes by me. I'll squash everything when putting this into master.

PR #192, opened 2020-10-23
Add test for LargeBasePrimitive

ssiccha:
Adds test for LargeBasePrimitive.
~~Moves two broken sporadic test cases.~~

### DIFF
--- a/tst/working/quick/PermLargeBasePrimitive.tst
+++ b/tst/working/quick/PermLargeBasePrimitive.tst
@@ -1,0 +1,13 @@
+#@local S19OnSets, jellyGroup, n, random, jellyGroupConj, ri
+
+# Test for LargeBasePrimitive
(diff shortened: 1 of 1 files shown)


ssiccha:
I removed the commit about the tests for sporadic group. If we merge #197, then we won't need that commit anymore.

fingolfin (changes requested):
This look good, except that neither the file content nor its name indicate that this is supposed to be a test for `LargeBasePrimitive`. I suggest to rename the file to match, and/or to mention `LargeBasePrimitive` in a comment

ssiccha:
The file is now called `PermLargeBasePrimitive.tst` and contains a comment. Also I changed the S_23 to be an S_19 which is quicker and also triggers `LargeBasePrimitive`.

ssiccha:
The tests that fail are sporadic groups that aren't recognized correctly. So from my POV this can be merged in.

ssiccha:
@fingolfin could you have a look again?

PR #193, opened 2020-10-23
Improve LargeBasePrimitive

ssiccha:
Fixes a comment in perm/largebase.gi and improves LargeBasePrimitive.

### Improve LargeBasePrimitive

Documents explicitly that LargeBasePrimitive can not handle fixed points.

Moves the NkrGetParameters call to the front, so that LargeBasePrimitive
can exit as soon as it knows that it is not applicable.

If LargeBasePrimitive is successfull, then it creates an isomorphism. So
there is nothing to do for the kernel.

### DIFF
--- a/gap/perm/largebase.gi
+++ b/gap/perm/largebase.gi
@@ -377,6 +377,11 @@ RECOG.AllJellyfish := function( G )
           s, l, params, p, n, k, r, jellyfishone, alljellyfish;
 
     N := LargestMovedPoint(G);
+    params := RECOG.NkrGetParameters( N );
+    if IsEmpty(params) then
+        return false;
+    fi;
+
     D := [ 1 .. N ];
 
     gens := ShallowCopy( GeneratorsOfGroup(G) );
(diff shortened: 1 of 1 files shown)


gap/perm/largebase.gi
```diff
@@ -507,6 +513,8 @@ FindHomMethodsPerm.LargeBasePrimitive :=
     res := RECOG.AllJellyfish(grp);
     if res = fail then
         return TemporaryFailure;
+    elif res = false then
+        return NeverApplicable;
```
fingolfin, line 517: How about changing `RECOG.AllJellyfish` to return `TemporaryFailure` resp. `NeverApplicable`, too? IMHO that'd make it easier to understand the code. Then above could become this:
```suggestion
    if not IsList(res) then
        return res;
```

fingolfin (commented):
Looks good, though I'd suggest one tweak. And perhaps we should first finish #192 then merge that, then update this PR to make sure we still pass the new test.

gap/perm/largebase.gi
ssiccha: Done.

ssiccha:
I've added documentation to 163a12f that LargeBasePrimitive may also return `NeverApplicable`.

ssiccha:
Compiling the manuals failed because some `apt update` failed. I've restarted the jobs.

fingolfin (approved):

Issue #157, opened 2019-07-15
Needed: (constructive) recognition of sporadic groups


fingolfin:
Thankful, Rob Wilson provided some valuable hints for this via private communication, which I paraphrase here (any mistakes are thus mine)

1. Some of the required functionality for recognition of sporadic groups is already available via the AtlasRep GAP package. Indeed, recog already uses some of these, but probably could and should use more.

2. The "Ryba algorithm" is a black box (with order oracle) algorithm for reducing membership testing to three instances of membership testing in involution centralisers. Could be implemented directly in GAP by working from the paper [HLO+08] (Holmes et al. JGT 11 (2008), 747-763). Alex Ryba used this method for sporadic group membership
testing many years earlier than this paper. 

3. A collection of hints for finding a stabiliser chain. The stabiliser chain method is a matrix group method, for efficient construction of base-and-strong-generating-set machinery. It was never written up for publication, but there are notes by Eamonn and Rob written at the time. Rob made them available to @hulpke, @ThomasBreuer and myself, and might be willing to share them with others if asked.

For the latter point, I wonder if it related to issue #19, and thus the content of [hints.gi](https://github.com/gap-packages/recog/blob/4d727563bdca73772ca96b66a02edff63b1c063e/gap/almostsimple/hints.gi) as well as the function `DoHintedStabChain` at <https://github.com/gap-packages/recog/blob/4d727563bdca73772ca96b66a02edff63b1c063e/gap/almostsimple.gi#L111-L203>

PR #195, opened 2020-10-23
Fix test and improve parameters for recognizing sporadics

ssiccha:


fingolfin:
Unfortunately, there is still another error:
```
working/quick/Sporadic.tst:43
# Input is:
TestSporadic("Suz");
# Expected output:
[ "Suz" ]
# But found:
[ "Fi22.2" ]
```

ssiccha:
Yes, and I feel I see the Suzuki error more often than in the past. :/

ssiccha:
Just to be sure I double checked the `Sporadic.tst` with an older version of master, namely with commit f5ca1f3. There we also have various regular errors.

ssiccha:
I've played around a little with the parameters of the `SporadicsByOrders` function. On my machine now running `Sporadic.tst` both before and after the change takes about 2 seconds. But the accuracy became a lot better.

Of course this isn't anything final but I think it's a good improvement from where we are now.

ssiccha:
On my machine, now all tests for sporadic groups work again.

gap/projective/almostsimple.gi
```diff
@@ -1328,7 +1328,7 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
   ordersseen := [];
   count := 0;
   gens := GeneratorsOfGroup(G);
-  limit := 120+Length(gens);
+  limit := 220+Length(gens);
```
fingolfin, line 1331: But where do these magic numbers come from? Are they based on a paper, or did someone just pull them from their a** ?

I'll merge it but I don't like it. Perhaps working on #157 will allow us to improve things.

fingolfin (approved):

tst/working/veryslow/Sporadic.tst
```diff
@@ -0,0 +1,15 @@
...
+# TODO: right now, we only test the first representation stored in AtlasRep;
+# ideally, we should test more
+gap> TestSporadic := function(name)
+>     local g, ri;
+>     g := AtlasGenerators(name,1).generators;
+>     g := Group(g);
+>     ri := EmptyRecognitionInfoRecord(rec(),g,IsMatrixGroup(g));
+>     return FindHomMethodsProjective.SporadicsByOrders(ri,g:DEBUGRECOGSPORADICS);
+> end;;
+
```
fingolfin, line 11:
```suggestion

#
```

gap/projective/almostsimple.gi
ssiccha: I just increased the values until SporadicsByOrders didn't rule out the wrong or too few groups. 

SporadicsByOrders stores for each sporadic group the most common element orders and their probabilities. During recognition of a group `G`, if after enough tries a common element order of a sporadic group `T` was not seen at all, then it is ruled out that `G` is isomorphic to `T`. For some types of sporadic groups that seems to happen too often. So maybe just some of the probabilities are incorrect. I can start by looking into the most common orders of the Suzuki group "Suz".

The only reference I found inside `recog/almostsiple.gi` that could maybe help here is `<Cite Key="KS09"/>` i.e. "Large element orders and the characteristic of Lie-type simple groups" by Kantor and Seress. Maybe it gives hints or references on how to choose the parameters to do the sporadic case.

fingolfin:
The CI tests now fail in another sporadic test

ssiccha:
On my machine the tests work fairly well. Like I can run the quick test suite 20 times without it breaking once. Weirdly though on travis naming of M24 failed three times in a row... I'm discarding the sporadic group commits so that we can merge this and repair the `ThrowAwayFixedPoints.tst`.

About the sporadic groups. I haven't found a paper explaining how the code in recog works. But I had a look into how magma does it and their approach is completely different and looks really simple. So I suggest to use their code and ask what they based it on. Of course, if someone finds an explanation of why the current code should work I'd also be happy to continue to use it. 

So here is roughly what magma does. In version 2.24-8 they use a function called `RecognizeSporadic` which computes a set
```
   maxords := {i : i in orders |
               #{j: j in orders | j mod i eq 0} eq 1};
```
and then e.g. does
```
   if maxords subset {6, 8, 11, 14, 15, 23} 
      and {11,14,15,23} subset maxords then
      Append (~spors, <18, 4, "M23">);
```
to find out if the group may be an M23.

PR #197, opened 2020-10-27
Replace SporadicByOrders by NameSporadic

### LINKED ISSUE #196: Implement sporadic naming as in magma
See the discussion in #195. Implementing the `RecognizeSporadic` function from magma is straight-forward. In magma it is called with the orders of at least 500 random group elements. So it might be a little bit slower then `SporadicsByOrders` which we currently use and which uses about 120 random elements. If the magma function turns out to be a lot more accurate though, we could get rid of `SporadicsByOrders`.

ssiccha:
Fixes #196.


ssiccha:
Unfortunately, running the old version of the `Sporadic.tst`, that is the version without the tests that were previously broken, but with excluded extensions like `Fi22.2`, now takes 13 seconds instead of 2 seconds. Though now the results should be correct more often. :-)

I'm running 50 iterations of the `Sporadic.tst` on my machine now. Let's see how many runs fail.

ssiccha:
If we want to merge this I can remove the `SporadicsByOrders` function and all of its, not otherwise used, subfunctions.

ssiccha:
Out of the 50 iterations, all passed on my machine.

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1429,126 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
   od;
   return false; # FIXME: false = NeverApplicable here really correct?
 end;
+
+#! @BeginChunk NameSporadic
+#! This method returns a list of sporadic simple groups that <A>ri</A>
+#! could be. It does not recognise the monster and the baby monster group. It
+#! is based on the magma v2.24.10 function <C>RecognizeSporadic</C>.
```
fingolfin, line 1436:
```suggestion
#! is based on the Magma v2.24.10 function <C>RecognizeSporadic</C>.
```

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1429,126 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+FindHomMethodsProjective.NameSporadic := function(ri)
+    local orders, setOfOrders, maximalOrders, counter, namesOfPossibleSporadics, i, j;
+    orders := [];
+    #do we need SetPseudoRandomStamp?
+    for i in [1 .. 500] do
+        Add(orders, RandomElmOrd(ri, "NameSporadic", false).order);
+    od;
+    # Compute maximal orders. Maximal in the sense that it does not divide the
+    # order of another group element.
+    setOfOrders := Immutable(Set(orders));
```
fingolfin, line 1446:
```suggestion
    setOfOrders := AsSet(orders);
```
First off, `Immutable` makes a copy of the set; you'd want `MakeImmutable` to avoid that. But best is to use `AsSet` which by definition directly returns an immutable set.

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1429,126 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+        counter := 0;
+        for j in setOfOrders do
+            if j mod i = 0 then
+                counter := counter + 1;
+            fi;
+        od;
+        if counter = 1 then
+            Add(maximalOrders, i);
+        fi;
+    od;
```
fingolfin, line 1458: This loops is somewhat inefficient, although it might not matter if `setOfOrders` is always small: if there are N orders, we perform N^2 divisions and checks. But clearly N over 2 suffice, so about a factor 2 in performance could be gained. Moreover, one can do away with counter. Something like this might work (untested):
```suggestion
    for i in [1..Length(setOfOrders)] do
        isMaximal := true;
        for j in [i+1..Length(setOfOrders)] do
            if j mod i = 0 then
                isMaximal := false;
                break;
            fi;
        od;
        if isMaximal then
            Add(maximalOrders, i);
        fi;
    od;
```

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1429,126 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+    namesOfPossibleSporadics := [];
+    if maximalOrders = [5, 6, 8, 11] then
```
fingolfin, line 1460: OK I am getting into micro optimizations here, but let me point out that this code will create a new list `[5,6,8,11]` each time it is executed. So if we call this function a 1000 times, we create 1000 such lists, which all need to be garbage collected later on. We could refactor the case perhaps by adding a global structure like this:
```gap
BindGlobal("SporadicsData", MakeImmutable([
  # entries:  [ name,  superset, subset ]
  [ "M11", [5, 6, 8, 11], [5, 6, 8, 11] ],
  ....
]));
And then the code can look like this:
```gap
  for sd in SporadicsData do
    if IsSubset(sd[2], maximalOrders) and IsSubset(maximalOrders, sd[3]) then
        Add(namesOfPossibleSporadics, sd[1]);
     fi;
  od;
```
Or perhaps `SporadicsData` should be a list of *records*, that might be easier to debug and gives more flexibility
```gap
BindGlobal("SporadicsData", MakeImmutable([
  rec( name:="M11", orders := [5, 6, 8, 11] ),
  ....
  rec( name:="M23", orders := [8, 10, 11, 12, 14, 15, 21, 23], required := [11, 21, 23] ),
  ....
]));
...
  for sd in SporadicsData do
    if not IsSubset(sd.orders, maximalOrders) then continue; fi;
    if IsBound(sd.required) then
      if not IsSubset(maximalOrders, sd.required) then continue; fi;
   else
    if Length(maximalOrders) <> Lenght(sd.orders) then continue; fi;
   fi;
   Add(namesOfPossibleSporadics, sd[1]);
  od;
```

tst/working/quick/Sporadic.tst
```diff
@@ -44,19 +44,13 @@ gap> TestSporadic("Suz");
 [ "Suz" ]
 gap> TestSporadic("Fi23");
 [ "Fi23" ]
-gap> TestSporadic("M12.2");
-[ "M12.2" ]
```
fingolfin, line 48: So it seems these extensions are not dealt with by the new code. How does Magma deal with them, then?

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1429,126 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
   od;
   return false; # FIXME: false = NeverApplicable here really correct?
 end;
+
+#! @BeginChunk NameSporadic
+#! This method returns a list of sporadic simple groups that <A>ri</A>
+#! could be. It does not recognise the monster and the baby monster group. It
```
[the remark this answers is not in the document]
ssiccha: Oops. ^^

gap/projective/almostsimple.gi
ssiccha: I'll try the record approach.

gap/projective/almostsimple.gi
ssiccha: I'm adding this. j and i have to be setOfOrders[j] or ..[i] though.

tst/working/quick/Sporadic.tst
ssiccha: We're now discussing this in #196.

ssiccha:
I still need to actually use `NameSporadic` in the recognition of leaves.

ssiccha:
I forgot to update the `local` statement.

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1430,114 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
   od;
   return false; # FIXME: false = NeverApplicable here really correct?
 end;
+
+# Data for the function NameSporadic
+BindGlobal("NameSporadicData", MakeImmutable([
+    rec(name := "M11",
+        maximalOrdersOverset := [5, 6, 8, 11],
+        length := 4),
```
fingolfin, line 1438: What's `length` for? It doesn't seem to be used?

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1430,114 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+                isMaximal := false;
+                break;
+            fi;
+        od;
+        if isMaximal then
+            Add(maximalOrders, setOfOrders[i]);
+        fi;
+    od;
+    namesOfPossibleSporadics := [];
+    return namesOfPossibleSporadics;
```
fingolfin, line 1542: Something seems to be missing here... :-)

gap/projective.gi
```diff
@@ -398,10 +398,10 @@ AddMethod(
 AddMethod(
     FindHomDbProjective,
     rec(
-        method := FindHomMethodsProjective.SporadicsByOrders,
+        method := FindHomMethodsProjective.NameSporadic,
         rank := 820,
-        stamp := "SporadicsByOrders",
-        comment := "generate a few random elements and compute the proj. orders",
+        stamp := "NameSporadic",
+        comment := "generate maximal orders",
```
fingolfin, line 404: Perhaps we should leave the old `AddMethod` in, albeit commented out, to make switching between the two implementations easy (while we are debugging things).

gap/projective/almostsimple.gi
```diff
@@ -1429,3 +1430,135 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+        return namesOfPossibleSporadics;
+    fi;
+    for name in namesOfPossibleSporadics do
+        Info(InfoRecog, 2, "Trying hint for ", name,
+             "...");
+        res := LookupHintForSimple(ri, Grp(ri), name);
+        if res = true then return Success; fi;
+        Info(InfoRecog, 2, "This did not work.");
+    od;
+    return false; # FIXME: false = NeverApplicable here really correct?
```
fingolfin, line 1563: Well, the only reason why it might succeed next time is because we missed some orders that excluded a potential sporadic... So in theory retrying might sometimes help, but only if we really didn't compute enough random elements. But doing this often would kill performance: if the group really is not sporadic then trying this again and again is useless.

fingolfin, line 1563:
```suggestion
    return NeverApplicable; # FIXME: NeverApplicable here really correct?
```

gap/projective/almostsimple.gi
ssiccha: Oh, it's a mistake that that isn't used. For M11, M12, M22 the maximalOrders set has to be equal to maximalOrdersOverset. I'm solving this by doing 
`maximalOrdersSubset := ~.maximalOrdersOverset),` for those groups

gap/projective/almostsimple.gi
ssiccha: You're right. I'm changing this to `return NeverApplicable;`.

gap/projective.gi
ssiccha: Ok. I'm also going to give `NameSporadic` a rank such that it is called more or less directly before `ThreeElementOrders`. `SporadicsByOrders` only generated 100 random elements or so, but `NameSporadic` generates 500 random elements, which sounds like it might be expensive.

ssiccha:
Rebased and finished. Let's see whether the tests pass.

ssiccha:
Tests now pass on my machine.

ssiccha:
Rebased onto master and enabled sporadic tests again. @fingolfin I didn't squash the changes yet so that it is easier for you to see my latest changes. Except for that I this PR is ready from my POV.

gap/projective/almostsimple.gi
```diff
@@ -1433,3 +1433,137 @@ FindHomMethodsProjective.SporadicsByOrders := function(ri,G)
...
+        if IsSubset(maximalOrders, data.maximalOrdersSubset)
+                and IsSubset(data.maximalOrdersOverset, maximalOrders) then
+            Add(namesOfPossibleSporadics, data.name);
+        fi;
+    od;
+    # HN works a bit differently
+    if IsSubset([9, 12, 14, 19, 21, 22, 25, 30, 35, 40], maximalOrders)
+            and (IsSubset(maximalOrders, [19]) or IsSubset(maximalOrders, [21]))
+    then
+        Add(namesOfPossibleSporadics, "HN");
```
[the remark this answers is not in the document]
ssiccha: I've pushed a commit which now uses `RECOG.NameSporadicData` to do the tests.

ssiccha (commented):
I added two commits. Once the PR is finished, we can squash everything into one commit.

.ctags
```diff
@@ -0,0 +1,21 @@
...
+--regex-GAP=/#Include *Label *= *"([a-zA-Z0-9_]+)"/\1/d,documentation/
+--regex-GAP=/DeclareCategory *\( *"([a-zA-Z0-9_]+)"/\1/c,category/
+--regex-GAP=/DeclareRepresentation *\( *"([a-zA-Z0-9_]+)"/\1/r,representation/
+--regex-GAP=/DeclareInfoClass *\( *"([a-zA-Z0-9_]+)"/\1/i,infoclass/
+--regex-GAP=/DeclareAttribute *\( *"([a-zA-Z0-9_]+)"/\1/a,attribute/
+--regex-GAP=/DeclareProperty *\( *"([a-zA-Z0-9_]+)"/\1/p,property/
+--regex-GAP=/DeclareSynonym *\( *"([a-zA-Z0-9_]+)"/\1/s,synonym/
+--regex-GAP=/DeclareSynonymAttr *\( *"([a-zA-Z0-9_]+)"/\1/s,synonym/
+--regex-GAP=/DeclareGlobalFunction *\( *"([a-zA-Z0-9_]+)"/\1/g,gfunction/
+--regex-GAP=/DeclareOperation *\( *"([a-zA-Z0-9_]+)"/\1/o,operation/
```
fingolfin, line 21: I don't mind adding a ctags file, but let's do it in a separate PR, or at least in a dedicated commit inside this PR?

ssiccha: Oh, I added that on accident. I have no idea how that got there. Weird. I've removed the file from the last commit.

ssiccha:
@fingolfin Could you have a look again? From my POV this is ready to be squashed and merged.

fingolfin (approved):
Looks good to me, thanks! Feel free to merge/squash.

PR #199, opened 2020-10-29
Classical.gi: Some micro optimizations

fingolfin:


fingolfin:
Apparently `KroneckerFactors` (part of non-constructive classical groups recognition) is never ever triggered by our test suite. Perhaps @aniemeyer has an idea how to construct examples that might trigger it?

ssiccha (commented):
LGTM

fingolfin:
Codecov complains because the code in `KroneckerFactors` is changed but not tested. We asked @aniemeyer about how to find test cases for this, and she told us this (translation by deepL and me, so any awkwardness or mistakes are not due to Alice, but my fault):
> This function is used for non-generic Orthogonal Plus groups in dimension 4 for q=8 or q>11.
> 
> All the non-generic cases are not really tested. What must happen for all these groups is that both the classical groups as test examples are inserted (and then the result should be true) as well as representatives of all maximum subgroups (and then the result should be false).

fingolfin:
@ssiccha we should try to see if we can find at least one test input to trigger this function

tst/working/quick/classical.tst
```diff
@@ -0,0 +1,20 @@
...
+true
+
+# brute force test
+gap> G := SL(2,5);;
+gap> SetX(G, G, function(g,h)
+>   local gh;
+>   gh:=KroneckerFactors(KroneckerProduct(g,h));
+>   return gh = false or (IsEqualProjective(gh[1], g) and IsEqualProjective(gh[2], h));
+> end);
+[ true ]
```
fingolfin, line 20: Always good to have dedicated tests for isolated functions like `KroneckerFactors` ! This promptly revealed a bug in the original version of this PR... oops.

Also makes for a handy benchmark. The above `SetX` call takes 650ms with this PR, down from 960ms before the PR. Not crucial but nice.

ssiccha (approved):

PR #200, opened 2020-10-31
Classical natural tests

ssiccha:


tst/working/quick/sl3.tst
```diff
@@ -1,11 +0,0 @@
...
->   [ [ Z(25)^20, Z(25)^8, Z(25)^23 ],
->     [ Z(25)^20, Z(25), Z(25)^16 ],
->     [ Z(25)^21, Z(25)^23, Z(25)^13 ] ],
->   [ [ Z(25)^16, Z(25)^0, Z(25)^4 ],
->     [ Z(25)^14, Z(25)^4, Z(25)^19 ],
->     [ Z(25)^23, Z(25)^17, Z(25)^12 ] ]
-> ]);;
-
-#
-gap> ri := RECOG.TestGroup(G, false, 152334000000);;
```
fingolfin, line 11: This test should be kept here, in the "quick" section, so its coverage is not reduced. If you dislike the duplicate, remove the copy in the "slow" test suite.

fingolfin (changes requested):

tst/working/quick/sl3.tst
ssiccha: I have instead made it so more of the tests from the `ClassicalNatural.tst` are now run in the `quick/` folder. I used `GL`s though I think that's no problem. For `GL`s we use the determinant homomorphism, then we recognize the `SL` in its kernel.

ssiccha:
Unfortunately on my machine, some of the smaller GLs fail. relatively regularly. I'll try to find out which ones that are. I have seen that there already are related issues, so I'll look there too.

fingolfin:
We now have `tst/working/quick/ClassicalNaturalNaming.tst`. Might be better to extend that instead.  Perhaps also get rid of `tst/testclassicalnatural.g`; for that, we maybe should...
- make the number of repetitions in `TestNaming` an argument??
- teach `TestNaming` to use randomize generating sets (but really, even without that we already find plenty bugs)
- add more input tests; inspired by `tst/testclassicalnatural.g` and this PR?

PR #201, opened 2020-11-09
Get rid of slpnice

ssiccha:


gap/perm/giant.gi
```diff
@@ -744,7 +744,6 @@ FindHomMethodsPerm.Giant :=
         SetIsRecogInfoForAlmostSimpleGroup(ri,true);
     fi;
     SetNiceGens(ri,StripMemory(res.gens));
-    Setslptonice(ri,res.slpnice);
```
fingolfin, line 747: Maybe apply the same changes to `gap/matrix/slconstr.gi` where the same pattern occurs, twice -- then afterwards, all occurrences of `slpnice` are eliminated.

ssiccha: Nice. Will do!

fingolfin:
This looks good now, but let's first get the CI working again before we merge other things.

PR #214, opened 2020-11-26
Fix naming of non-generic orthogonal groups

fingolfin:


tst/working/quick/nongeneric-classical.tst
```diff
@@ -0,0 +1,17 @@
+# test some non-generic cases
```
ssiccha, line 1: What about calling this file `NonGenericClassical.tst`? The other test files also use camel case for their names.

fingolfin: also fine by me.

Alice just pointed out to me that we should also test (representatives of) maximal subgroups to make sure they are *not* "recognized" incorrectly as their parent.

gap/matrix/classical.gi
```diff
@@ -759,26 +759,26 @@ RECOG.TestRandomElement := function (recognise, grp)
             Info(InfoClassical,2,o1, " ", o2);
             ### ACN March 2007 needed Projective order
             if  q mod 2 = 0 and (o1 <= 2 or o2 <= 2) then
-                return fail;
+                return TemporaryFailure;
             elif q mod 2 <> 0 and (o1 <= 4 or o2 <= 4) then
-                return fail;
+                return TemporaryFailure;
```
fingolfin, line 764: @aniemeyer we need your input here... does the above change seem plausible. The old value `fail` / `NeverApplicable` means "give up, we have proven this can't be such-and-such group, while `TemporaryFailure` means it may be tried again later.

We couldn't see anything in your papers which would support that we abort here; and with is change, the non-generic SO(+1,4,q) get named much more reliably.

ssiccha: I think false meant "we have proven that this can't be a classical group". So the lines where this PR actually changes behavior are the ones where false is replaced by temporary failure.

fingolfin: Yeah I commented on the wrong line (once again proving that false/fail were a bad choice as return values

gap/matrix/classical.gi
```diff
@@ -759,26 +759,26 @@ RECOG.TestRandomElement := function (recognise, grp)
...
                     fi;
                     if not ForAny( PrimeDivisors(o2), r ->
                       IsPrimitivePrimeDivisor(p,2*recognise.a,r)) then
-                        return false;
+                        return TemporaryFailure;
                     fi;
                 elif q = 8 then
                     if not o1 mod 9 = 0 or not o2 mod 9 = 0 then
-                        return false;
+                        return TemporaryFailure;
```
fingolfin, line 778: @aniemeyer so here for example: we are looking for two elements of order 9; if only find one, and the other is of order 3, then in the previous version of the code we ended here with `false = NeverApplicable` and so recognition overall failed. That seems wrong; with the changes in this PR, more naming tests pass. The only drawback I can think of is "we might try too hard in some cases", but that's it. Can you confirm?

aniemeyer (approved):

PR #216, opened 2020-12-03
Improvements to three element orders

ssiccha:
## Rename twoelorders.gi -> threeelorders.gi and improve its source comment.
The source comment now explains why RECOG.grouplist only stores two
largest element orders.

## Fix source comment of ThreeLargeElOrders

It claimed that ThreeLargeElOrders could also handle sporadic groups.
However, neither does the paper [KS09] which describes the algorithm
mention sporadic groups, nor does did I find anything relating to
sporadic groups in the function.

I tested this by taking matrix representation via AtlasGenerators for
M11, Co2 and J2 and then calling ComputeSimpleSocle and
ThreeLargeElOrders. ThreeLargeElOrders returned fail in all three cases.

## Improve explanatory string of ThreeLargeElOrders

Say that it recognises lie groups and computes their characteristic.

### DIFF
--- a/gap/projective.gi
+++ b/gap/projective.gi
@@ -461,7 +461,7 @@ AddMethod(
         method := FindHomMethodsProjective.ThreeLargeElOrders,
         rank := 500,
         stamp := "ThreeLargeElOrders",
-        comment := "look at three large element orders",
+        comment := "recognise Lie type groups and get its characteristic",
     )
 );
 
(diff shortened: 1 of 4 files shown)


ssiccha:
I've also added another commit which states the prerequisuites of ThreeLargeElOrders.

fingolfin (approved):

PR #217, opened 2020-12-04
Fix naming of non-generic unitary groups

fingolfin:


gap/matrix/classical.gi
```diff
@@ -1440,8 +1440,6 @@ RECOG.NonGenericUnitary := function(recognise, grp)
         fi;
          Info(InfoClassical,2,"group contains SU(",
               recognise.d, ", ", recognise.q, ");");
-        recognise.isSpContained := true;
```
fingolfin, line 1443: @aniemeyer can you think of any reason other than "a copy & paste error)" why this line was there? I.e. is me removing it likely to be breaking something??

aniemeyer (approved):

PR #222, opened 2020-12-09
Minor clean-up

ssiccha:


fingolfin (commented):
What does this "clean up"? What was "untidy" before?

ssiccha:
First one adds `F` to the `depthstring`, then one recursively calls `RecogniseGeneric` on the factor, then `F` is removed from `depthstring` again.

So I thought it would be nicer to have the `Add` and `Remove` statements both directly around the `RecogniseGeneric` call, instead of doing `ValidateHomomInput` in-between.

fingolfin (approved):

PR #223, opened 2020-12-09
Changes in Unitary and Orthogonal from Summerschool 2019 by Cheryl an…

aniemeyer:
…d Alice


fingolfin:
Sporadic test failures due to TestSporadic *sigh* @ssiccha I am tempted to say we should just disable all of those for now until your replacement code lands (i.e. that PR should add them back)

gap/matrix/classical.gi
```diff
@@ -1712,11 +1723,25 @@ RECOG.NonGenericOrthogonalPlus := function(recognise,grp)
              return false;
         fi;
     elif d = 8 and  q =  5 then
-        #ACN Feb 07: added fix - need also elements of order a mult of 3
-        #            these are missing in the paper [NP99]
-        if not HasElementsMultipleOf( recognise.orders, [7,13,3])  then
+        ## 2.July.2019: There is a mistake in the paper here
+        ## There are maximal subgroups of Omega+(8,5) that contain 
+        ## elements of order 7,13,3
+        if not HasElementsMultipleOf( recognise.orders, [7,13,3,312])  then
```
fingolfin, line 1729: So where does the 312 come from? It's not a prime, and I just had a look at POmega(8,5) and based on a random sampling it only contains elements of order 156, not 312. 

Wouldn't 31 make more sense?

gap/matrix/classical.gi
```diff
@@ -1701,22 +1707,36 @@ RECOG.NonGenericOrthogonalPlus := function(recognise,grp)
...
-        #            these are missing in the paper [NP99]
-        if not HasElementsMultipleOf( recognise.orders, [7,13,3])  then
+        ## 2.July.2019: There is a mistake in the paper here
+        ## There are maximal subgroups of Omega+(8,5) that contain
+        ## elements of order 7,13,3
+        if not HasElementsMultipleOf( recognise.orders, [7,13,3,312])  then
             return fail;
         fi;
+
+        pgrp := ProjectiveActionOnFullSpace( grp, recognise.field, d );
```
fingolfin, line 1728: On my laptop I get:
```
gap> ProjectiveActionOnFullSpace( Omega(+1,8,5), GF(5), 8);; time;
648
```
So this takes more than half a second. That seems rather slow. We also may end up re-computing this action later on... As it is, it caused a severe slowdown of our test suite, because now testing with the input `O(+1,8,5)` is very slow. Before this PR:
```
gap> ReadPackage("recog", "tst/naming.g");
gap> TestNaming("SO", +1, 8, 5); time;
645
```
With this PR:
```
gap> ReadPackage("recog", "tst/naming.g");
gap> TestNaming("SO", +1, 8, 5); time;
212476
```

So it goes in total from 0.6 seconds to 3.5 minutes!!!

Do we really need the whole orbit?  Perhaps we can instead just check the length of the orbit of a single vector or so?

ssiccha: I just noticed that we do the same computation in other cases too. So is this slower because the dimension is bigger or do we maybe not test the other cases?

fingolfin: The other cases are indeed significantly smaller; just computing the orbit takes some time. Moreover 5^7 > 2^16 so "big" permutations have to be used; 4*5^7 = 320kb per permutation kills most caches quickly.

ssiccha: @aniemeyer and I are looking at this.

I'll use magma to compute the maximal subgroups of GO+(8,5). Then we can see how to distinguish them from the Omega+(8,5).

ssiccha: Also, we should write down which theorem this code is based upon.

ssiccha:
I added some "work in progress" code how we are fixing Omega+(8,5). We're not finished yet though.

gap/matrix/classical.gi
ssiccha: Some thoughts:
- It should be possible to compute only one orbit. I've tried it and unfortunately that also takes half a second. 
- I haven't figured out yet how to compute the maximal sugroups of GO+(8,5) in Magma. I can get the maximal ones of Omega+(8,5), but that's it.

fingolfin: Do you know by now what the relevant "bad" maximal subgroup is?

Also, I noticed that while I complained that `pgrp := ProjectiveActionOnFullSpace(...)` takes half a second, the call `Size(pgrp)` later actually takes *5 seconds* ... So yeah, we need to work on this :-)

fingolfin:
```
gap> orb:=Orbit(G,One(G)[1], OnLines);; time;
90
gap> Length(orb);
19656
gap> pg:=Action(G,orb,OnLines);; time;
223
gap> Size(pg2); time;
8911539000000000000
1369
```

So perhaps this would work:
```gap
orb:=Orbit(grp, One(grp)[1], OnLines);
if not Length(orb) in [ 19656, 39000 ] then  ... return false; fi
pgrp:=Action(G,orb,OnLines);
if Size(pgrp) ...
```
Of course `pgrp` may underestimate the actual size of the group. In order to understand whether the above is sufficient or not, I'd need to better understand what the group is we want to distinguish. @aniemeyer just told me that it is O(7,5) which is of course muuuch smaller; it has size 914004000000000. So perhaps it suffices to test whether `grp` resp. `pgrp` has more than 914004000000000 elements; for this, one could abort the random schreier sims early as soon as one has a group bigger than that.

ssiccha: @fingolfin Oh, interesting! Aborting the schreier sims computation then seems like the way to go. Cheryl and Alice had a chat about how to do this case last week (I think) and we'll try to put it into code after the GAP days.

ssiccha:
Here is our Hack.md. I haven't updated it for some time:
https://hackmd.io/ZwPI13VVQQuvdG1lSdMhCw

ssiccha:
Rebased onto master. I think I'll just go ahead and compare the gap and magma code for the classical groups on a case by case basis and see where they differ substantially. Since the magma code is fairly well tested, that might be good enough for us.

gap/matrix/classical.gi
```diff
@@ -1313,20 +1316,21 @@ function( recognise, grp )
             recognise.needForms :=  true;
             return fail;
         fi;
+        # TODO?
```
fingolfin:
```suggestion
```

gap/matrix/classical.gi
```diff
@@ -1592,11 +1606,12 @@ function(recognise, grp)
...
         (7*(q-1)) mod i <> 0 # order does not divide 7*(q-1)
          )) then
-            # 5 is not ppd of q^2-1
             return CheckFlag();
         fi;
+        return fail;
     elif d = 3 and q = 4 then
-        if Order(grp) mod 216 = 0 then
+        # Take the natural action of grp, and compute its order
+        if Order(Action(grp, recognise.field ^ d, OnPoints)) mod 216 = 0 then
```
fingolfin: I don't understand why this is different from the previous code, i.e.
```suggestion
        if Order(grp) mod 216 = 0 then
```

PR #228, opened 2020-12-10
PositionProperty -> ForAny/ForAll

fingolfin:


gap/matrix/classical.gi
```diff
@@ -2040,36 +2025,31 @@ RECOG.NonGenericOrthogonalCircle := function( recognise, grp )
         fi;
     elif d = 3 and q >=31 and IsPowerOfTwo(q+1) then
         s := Log2Int(q+1);
-        if PositionProperty(recognise.orders,
-            i->(i > 2 and (q-1) mod i = 0))=fail then
+        if not ForAny(recognise.orders, i-> i > 2 and (q mod i = 1)) then
```
FriedrichRober, line 2028: Why do we change `(q-1) mod i = 0` to `(q mod i = 1)` only in this line but not in other places in this file?
I would suggest to make this consistent.

PR #234, opened 2020-12-17
Enable TestGroupOptions.tryNonGroupElements

ssiccha:
This checks whether recog can correctly recognise, that a given input
does not lie in the recognised group.

-
Lets see whether the tests pass.


ssiccha:
There's only three test failures which looks pretty good.

PR #238, opened 2021-02-16
Update Codecov badge

ssiccha:


README.md
```diff
@@ -1,5 +1,5 @@
 [![Build Status](https://github.com/gap-packages/recog/workflows/CI/badge.svg)](https://github.com/gap-packages/recog/actions?query=workflow%3A%22CI%22)
-[![Code Coverage](https://codecov.io/github/gap-packages/recog/coverage.svg?branch=master&token=)](https://codecov.io/gh/gap-packages/recog)
+[![Code Coverage](https://app.codecov.io/github/gap-packages/recog/coverage.svg?branch=master&token=)](https://codecov.io/gh/gap-packages/recog)
```
fingolfin, line 2: What's the motivation?

As it is, it's not clear to me whether this change is desirable or advisable. For starters, I don't even know if those app.* links will keep working -- for all I know they are a temporary experiment and will be withdrawn again tomorrow?

And *if* we make that change, why only change the badge, but not the link destination, too?

ssiccha: Ah, then I misunderstood. I thought it would be the new design which will supersede the old one. Let me  just close this then.

fingolfin: Using the app. ones seems fine, see https://community.codecov.io/t/whats-the-difference-between-app-codecov-io-and-codecov-io/2563/2?u=fingolfin  -- but then you should also change the link, not just the image URL :-)

ssiccha:
I've fixed the link. ^^

ssiccha:
Hm, I can't reopen this PR since I force pushed this branch? :thinking: Well, I'll open a new one then.

PR #241, opened 2021-03-30
Move cyclic2.gi into example dir

ssiccha:
It looks like the file cyclic2.gi was put into `gap/examples` by accident. Or
is there a reason for it being in `gap/examples` instead of in `examples`?


fingolfin (approved):
Well, it's an example for a recognition method... but I don't mind where it sits. More important would be to finally integrate the tutorial for adding a recognition method into the recog manual! Then this example could be referenced or even included as part of the text.

PR #245, opened 2021-03-31
Rename Commute > AreCommutingNonProjectiveMatrices

ssiccha:


gap/projective/AnSnOnFDPM.gi
```diff
@@ -58,7 +58,7 @@ end;
 
 ## Functions for efficient matrix operations
 
-RECOG.Commute := function(g, h);
+RECOG.AreCommutingNonProjectiveMatrices := function(g, h);
```
fingolfin, line 61: Matrices are by default not projective, so this feels overly redundant.
```suggestion
RECOG.AreCommutingMatrices := function(g, h);
```
Also, one immediately wonders why this doesn't use your new AreCommuting/DoCommute methods. Presumably there is a reason -- if so, it should be stated in a comment. (My guess is: you are concerned this is called with "projective" groups)

Finally, why change this but not the Conj and ConjInv methods afterwards?

FriedrichRober: I agree. It is unclear why we shouldn't use the new `AreCommuting/DoCommute` method. I thought the idea was to use this the same way as the `isone` entry in the recog node in order to deal with projective groups.

ssiccha:
I'll try to remove the `ConvertToMatrixRep` calls. To make sure this does not slow down thinkgs, I'll check with asserts whether the input is in the correct representation.

ssiccha:
Actually, I have thought about this again, see #260. I'm closing this PR since the whole file might be deleted anyways.

PR #248, opened 2021-04-01
WIP doc: integrate howto write a recog method

ssiccha:
I'm putting this here for now so that it doesn't get lost.

### DIFF
--- a/temp.xml
+++ b/temp.xml
@@ -1,0 +1,53 @@
+TODO: where to put this?
+
+## A quick overview on how group recognition works
+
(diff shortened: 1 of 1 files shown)


fingolfin:
I am confused why this wasn't part of PR #250 and thus part of the HOWTO that was merged.

Yes, some of this could also be used to improve other parts of the manual. But the point here was to have it in the HOWTO specifically so that people have all the information immediately available, instead of having to first read an unknown set (possibly all) of the other chapters/sections of the manual...

fingolfin:
This seems to be a dump of the content of https://hackmd.io/UQHK8XZgSJyZrUMEzb_6eQ?view but I don't see how it helps us, so closing now.

PR #249, opened 2021-04-01
Merge documentation of CallMethods

ssiccha:


fingolfin (approved):
Honestly I have doubts whether this is the right direction on the long run (we might be better served with a section/chapter describing the overall system, instead of putting all of that into the documentation of a single function). But if you think it's an improvement right now, OK, why not...

fingolfin:
This needs to be rebased

ssiccha:
> we might be better served with a section/chapter describing the overall system, instead of putting all of that into the documentation of a single function

sounds good to me. I'll try to keep it in mind.

Issue #177, opened 2020-06-19
Improve names of functions, introduce uniform terminology


ssiccha:
This is project 14 from the [Summer School Matrix Group Recognition](https://lbfm-rwth.github.io/Summer-School-MGRP/projects/).

A script to rename function names in the declaration, implementation, test, and documentation files was written by me and @FriedrichRober and is available in the branch [Improve-Function-Names](https://github.com/ssiccha/recog/tree/Improve-Function-Names) at my fork.

 If anyone wants to use the script in that branch to rename functions you can:
- take the code from my fork
- perform the renaming
- commit the changes and then only keep the renaming commit by using e.g. `rebase --interactive` or `cherry-pick`

More details are given in the respective `RENAME.md` file.

PR #252, opened 2021-04-02
Improve doc of gap/matrix/slconstr.gi

ssiccha:
- add a header comment which of these functions is actually used
- document SLCR.FindHom
- reference SLCR.FindHom in the documentation of
  FindHomMethodsMatrix.NaturalSL

TODO: the documentation of SLCR.FindHom does not appear in the manual.
@fingolfin, could you point me to the file I need to change? I had a look but didn't find anything.

### DIFF
--- a/gap/matrix/slconstr.gi
+++ b/gap/matrix/slconstr.gi
@@ -14,6 +14,9 @@
 ##
 ##  Peter Brooksbank's code for constructive SL and PSL recognition.
 ##
+##  The function SLCR.FindHom can be used to find a constructive isomorphism
+##  from a group G to an SL(d,q), provided one knows d and q.
+##
 #############################################################################
 
 SLCR := rec();
@@ -2422,7 +2425,12 @@ end;
 ##end;
 
 # Here comes the code linking all this to the recog package:
-
+#! @BeginChunk SLCR.FindHom
+#! Given two integers d and g and a group G which must be isomorphic to SL(d,q),
+#! this function recognises G constructively and stores the result in ri.
+#!
+#! TODO: find a better name.
+#! @EndChunk
 SLCR.FindHom := function(ri,G,d,q)
   local GM,data,e,genlist,i,j,p;
   p := Factors(q);


ssiccha:
@danielrademacher Maybe this is also interesting for you.

ssiccha:
Also, `FindHom` should be renamed. While we're at it, what would be a good naming convention for leaf methods so that they can be easily grepped for?

danielrademacher:
Yes, that's great to know! Thanks! 
I have to check what function is used in Akos and Max code and to run some tests which functions are currently best in GAP for constructive recognition.

And I agree with Sergio, we should rename the function and choose clearer names to differentiate which function does what. The names should be chosen consistently in a certain way. Then it would be easier for users and programmers of the package. But that is probably connected with an enormous amount of effort. Shall we open an issue about this?

ssiccha:
@danielrademacher To in general come up with better names for many functions we already have an issue, see #177 and [our note with proposals for new names](https://hackmd.io/bsihJmkLQ06Sm-Zxv8oAWA?view) which we made during the summer school. For #177 @FriedrichRober and I only looked into the documented names though. Thanks for the comment, I'm looking into finding a few names from our proposal note to open a PR with. That has also been lying around for way too long.

Back to the `FindHom` function. That one was not documented before. So we didn't consider it for #177 yet. Also we had not thought about a naming scheme for **leaf** methods yet. I guess in the documentation we should also separate the splitting methods and the leaf methods.

ssiccha:
@danielrademacher See #253.

gap/matrix/slconstr.gi
```diff
@@ -2422,7 +2425,10 @@ end;
 ##end;
 
 # Here comes the code linking all this to the recog package:
-
+#! @BeginChunk SLCR.FindHom
+#! Given a group G which is isomorphic to SL(d,q), and integers d and q, this
+#! function recognises G constructively. TODO: find a better name.
```
fingolfin, line 2430:
```suggestion
#! Given a two integers d and g and group G which must be isomorphic to SL(d,q),
#! this function recognises G constructively and stores the result in ri.
#!
#! TODO: find a better name.
```

gap/matrix/slconstr.gi
```diff
@@ -2465,7 +2471,7 @@ SLPforElementFuncsMatrix.SLConstructive := function(ri,x)
 end;
 
 #! @BeginChunk NaturalSL
-#! TODO
+#! TODO: This function is not used at the moment, but SLCR.FindHom is used.
```
fingolfin, line 2474: Iissue #257 would also take care of this comment

ssiccha: I'm removing this comment and will do #257 and #258 next.

ssiccha:
Added suggestions and rebased.

danielrademacher (approved):
Looks good.

fingolfin (approved):

PR #253, opened 2021-04-02
Improve and unify names

ssiccha:
RecognitionInfoFamily -> RecogNodeFamily
IsRecognitionInfo -> IsRecogNode
RIFac -> ImageRecogNode
RIKer -> KernelRecogNode

PR 1 for #177

### DIFF
--- a/doc/recognition.xml
+++ b/doc/recognition.xml
@@ -52,7 +52,7 @@ with a technical description of the recursive procedure
 behind our main function <Ref Func="RecogniseGroup"/>.
 In Section <Ref Sect="rirecord"/>
 we describe the return type of <Ref Func="RecogniseGroup"/>
-which we call <Q>recognition info records</Q>.
+which we call <Q>recognition nodes</Q>.
 The methods to find homomorphisms are described
 in Section <Ref Sect="findhomo"/>.
 Finally,
@@ -65,7 +65,7 @@ recognition of different types of groups.<P/>
 At the heart of the recognition
 procedure is a function called <Ref Func="RecogniseGeneric"/> which gets
 a &GAP; group object and returns a
-so-called <Q>recognition info record</Q> (see Subsection
+so-called <Q>recognition node</Q> (see Subsection
 <Ref Subsect="rirecord"/> for details). Success or failure will be indicated
 by this record being in the filter <Ref Filt="IsReady"/> or not.<P/>
 To know how to find homomorphisms
@@ -80,7 +80,7 @@ To describe the algorithm executed by <Ref Func="RecogniseGeneric"/>
 we first summarise it in steps:
 
 <Enum>
-<Item>Create a new, empty recognition info record.</Item>
+<Item>Create a new, empty recognition node.</Item>
 <Item>Use the database of <C>FindHomomorphism</C> methods and the method
     selection procedure described in Chapter <Ref Chap="methsel"/> to
     try to find a homomorphism onto a smaller group or an isomorphism
@@ -89,12 +89,12 @@ we first summarise it in steps:
 <Item>If an isomorphism is found or a method somehow else recognises
     the group in question, such that we can write elements as straight
     line programs in the generators from now on, then make the recognition
-    info record a leaf of the recognition tree and return success.</Item>
+    node a leaf of the recognition tree and return success.</Item>
 <Item>Otherwise the function sets up all the data for the homomorphism and
     calls itself with the image of the homomorphism. Note that this
     might use another database of recognition methods because the
     homomorphism might change the representation of the group.</Item>
-<Item>After successful recognition of the factor group the procedure has
+<Item>After successful recognition of the image group the procedure has
     to recognise the kernel of the homomorphism. The first step for this
     is to find generators. If they are not already known from the
     <C>FindHomomorphism</C> method, they are created by producing random
@@ -119,8 +119,8 @@ group as a straight line program in the generators as follows:
 
 <Enum>
 <Item>Map the element through the homomorphism.</Item>
-<Item>Write the element in the factor group as a product of the nice
-    generators in the factor group.</Item>
+<Item>Write the element in the image group as a product of the nice
+    generators in the image group.</Item>
 <Item>Apply the resulting straight line program to the preimages of those
     nice generators and calculate the quotient, which will now lie in
     the kernel.</Item>
@@ -138,7 +138,7 @@ verification phase.<P/>
 The latter procedure to write elements as straight line programs in the
 generators is implemented in the function <Ref Func="SLPforElementGeneric"/>
 which will be called automatically if one calls the <Ref Func="SLPforElement"/>
-function of the resulting recognition info record (see
+function of the resulting recognition node (see
 <Ref Attr="slpforelement"/>).<P/>
 
 It is now high time to give you the calling details of the main recursive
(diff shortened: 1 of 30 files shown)


ssiccha:
Merging this may clash with #176. I'll try whether I can rebase that branch easily on top of this one. Otherwise we'd have to wait until #176 is merged first.

doc/recognition.xml
```diff
@@ -195,19 +195,19 @@ A recognition info record is a &GAP; component object. It is a member
...
 <ManSection>
-    <Filt Name="IsRecognitionInfo" Type="Category"/>
+    <Filt Name="IsRecogNode" Type="Category"/>
 </ManSection>
 
 and is <Ref Filt="IsAttributeStoringRep" BookName="Ref"/>, such that we can
 define attributes for it, the values of which are stored once they are
 known. A recognition info record always represents a whole binary tree of such
-records, see the attributes <Ref Attr="RIFac"/> and <Ref Attr="RIKer"/>
+records, see the attributes <Ref Attr="ImageNode"/> and <Ref Attr="KernelNode"/>
```
fingolfin, line 210: I don't like that that these names do not reference the fact that this is for recognition. If I here "image node" then I don't know what this is about at all.

Elsewhere in this PR you introduce  `IsRecogNode`. So how about calling these `ImageRecogNode` or `RecogNodeForImage` or so (and similar for the kernel)?

ssiccha: I like `ImageRecogNode`. 👍

ssiccha:
@FriedrichRober The tests fail because we rename e.g. `RIFac` but not `Has... ` and `SetRIFac`. We need to update our script accordingly.

doc/renaming_table.xml
```diff
@@ -0,0 +1,30 @@
...
+	<Item> <Ref Attr="ImageNode"/> </Item>
+</Row>
+<HorLine/>
+<Row>
+	<Item> RIKer </Item>
+	<Item> <Ref Attr="KernelNode"/> </Item>
+</Row>
+<HorLine/>
+</Table>
+</Chapter>
```
wilfwilson, line 30: Why is there a closing `Chapter` tag here?

doc/renaming_table.xml
```diff
@@ -0,0 +1,30 @@
+<Table Align="|l|l|">
```
wilfwilson, line 1: Do you intend to include `doc/renaming_table/xml` in the manual somewhere?

wilfwilson (commented):
In general I approve, while also supporting Max's suggestion.

This is perhaps an ignorant question (I've not really thought about recog for quite a while) - is naming this in terms of 'image' definitely more appropriate than naming it in terms of 'factor'? I only ask because (I presume) the old name `RIFac` was referring to a 'factor'; although the actual documentation talked about the 'image'.

doc/renaming_table.xml
ssiccha: Yepp. I haven't looked into the renaming code for ages and I forgot to add a few changes. :grin:

wilfwilson:
Cool 🙂 I'm happy for it to be called `image`, but as just curious.

ssiccha:
Rebased onto master, fixed the tests and the documentation. Please in particular have a look at the file `doc/renaming.xml` or Chapter 8 "Renaming".

doc/renaming_table.xml
```diff
@@ -0,0 +1,29 @@
...
+<Caption> Renaming</Caption>
+<HorLine/>
+<Row>
+	<Item> <E>Old Name</E> </Item>
+	<Item> <E>New Name</E> </Item>
+</Row>
+<HorLine/>
+<Row>
+	<Item> RecognitionInfoFamily </Item>
```
wilfwilson, line 10:
```suggestion
	<Item> <C>RecognitionInfoFamily</C> </Item>
```
and so on.

Also, perhaps index keys could be added for `RecognitionInfoFamily`, etc, so that the old names still appear in the index (but they now direct you to the correct place now, so `RecognitionInfoFamily` would take you to the documentation for `RecogNodeFamily`, etc). I _think_ this is popular with GAP Doc, but I can't remember how.

wilfwilson: Perhaps you have to add `<Index><C>RecognitionInfoFamily</C></Index>` to the documentation for `RecogNodeFamily`.

ssiccha: That works for me. Nice, thanks! Adding the old names to the index sounds like a good idea to me.

I think I would prefer it to have the index entries of the old names link to the renaming chapter instead. That way there can't be any confusion about why one clicks e.g. on `RecognitionInfoFamily`, but does not find any documentation for `RecognitionInfoFamily`. Especially since sometimes a new name @FriedrichRober and I came up with can be very different from the old name, if the latter was very cryptic.

wilfwilson: Yeah that's a better idea.

ssiccha: In order for such an entry to be sorted correctly into the index one also needs to specify a sort `Key` like this:
`<Index Key="RecognitionInfoFamily"><C>RecognitionInfoFamily</C></Index>`

FriedrichRober (approved):
I think the names are now very descriptive and the renamings are well documented.

fingolfin:
Regarding factor vs. image: of course both make sense. However, we do talk about the "factor group" in many comments and  the manual. There also still the attributes `methodsforfactor` or `forfactor`, which now don't match anymore. Of course w could also rename them.

Taking a step back, I guess both names make sense: a factor group is the image of a group homomorphism... However, my mind has a slight tendency to think of this as a "proper factor", i.e. the homomorphism has a non-trivial kernel, while in our case, there are also sometimes isomorphisms... While "image" does not have such connotation to me. So maybe that means "image" would be better indeed, at least if other people suffer from the same prejudice regarding "factor" as I do :-)

ssiccha:
Hm, interesting. When you think of an inner node in the recog tree as a homomorphism it is IMO more direct to talk of kernels and images. When you think of an inner node as a group, it is IMO more direct to talk of normal subgroups and factors. At the moment though we're mixing both by talking about kernels and factors.

And yes, when I hear or read "factor" I also don't think of a factor by the trivial subgroup. :D

ssiccha:
Two follow up PRs would be:
- turn `Info` into `Node` where sensible
- turn `Factor` into `Image` where sensible

In both cases we should also search the manual for left over occurences of `Info` and `Factor`. I can use my tags file to search for all names which contain `Info` or `Factor`.

wilfwilson:
I've enjoyed this discussion 🙂

ssiccha:
I've rebased this onto master and squashed the "fixup" commits. No other changes were introduced.

ssiccha:
I'll try to get #176 into a state where it can be merged into master. Then I'd merge in this PR (#253) and then @FriedrichRober and I can continue work on #176 in a different branch and PR.

ssiccha:
Rebased onto master. Now I need to change `factor` in the manual to `image` and make sure that updating the index is also done in the script @FriedrichRober and I wrote.

ssiccha:
I've replaced most occurences of `factor` with `image`. Of course, where `factor` meant a prime factor or the factor of a polynomial I left it alone. :)

ssiccha:
@fingolfin Could you have another look?

ssiccha:
TODO: I just saw, that `InstallMethod` calls often have a string `for recognition infos` or something similar. I need to adjust this.

ssiccha:
Ok, I got rid of almost all appearances of `'\<info\>'`. The remaining ones don't have anything to do with recog info records.

wilfwilson (approved):
LGTM.

ssiccha:
`bugfix.tst` failed because now line breaks are different.

PR #255, opened 2021-04-06
Tweak GitHub Actions

wilfwilson:
Just a couple of small changes to bring this into line with the example package.

Also, I don't think there's a need to compile the documentation on every test job, so I've removed that. It should save some seconds.

(I also don't think it's necessary to explicitly clone AutoDoc).


fingolfin (approved):

ssiccha:
Thanks @wilfwilson :)

PR #263, opened 2021-04-08
Add a RecogMethod type

### LINKED ISSUE #258: Add a method type
We could turn methods into records or types which could then store meta information like:
- its stamp
- its SLPForElement and Homoms check input #107
- whether it is a reduction / leaf method
- can handle perm / matrix / projective groups 
- bibliography

What else could be in there?

ssiccha:
Fixes #258.

Things to double-check before this gets merged:
- did I remove all the hacks?
- did I update the method HOWTO?


ssiccha:
When I'm done I must check whether I forgot to remove any hacks.

ssiccha:
I've got a first version of `RecogMethod`s working. I've introduced several hacks so that I'm able to incrementally change every method into a `RecogMethod` object. When I'm finished I'm going to remove them again.

ssiccha:
I like what I've done so far. However, before I start converting all methods to `RecogMethod`s, could you check whether you disagree with some of my design choices @fingolfin? I think you can ignore `methsel.gi` and `recognition.gi` for now since they contain mostly hacks.

gap/base/methods.gd
```diff
@@ -0,0 +1,40 @@
...
+
+#! @Chapter methsel
+#! @Section whataremethods
+#! @Description
+#! TODO: how to get this into the chapter with the GAPDoc label "methsel"?
+#! TODO
+#! method     : the function itself
+#! rank       : an integer rank
+#! stamp      : a string describing the method uniquely
+#! comment    : an optional comment to describe the method for humans
```
fingolfin, line 27: I assume the above comment is a draft and perhaps not meant for `IsRecogMethod` but rather for `RecogMethod`?

gap/base/methods.gi
```diff
@@ -0,0 +1,59 @@
...
+    Append(s, ViewString(FunctionOfRecogMethod(m)));
+    Append(s, ">");
+    return s;
+end);
+
+InstallGlobalFunction(RecogMethod,
+function(stamp, comment, func)
+    local r;
+    r := rec();
+    Objectify(RecogMethodType, r);
```
fingolfin, line 49: Perhaps use `ObjectifyWithAttributes`, it can have better performance

gap/base/methods.gd
```diff
@@ -0,0 +1,40 @@
...
+BindGlobal("RecogMethodType",
+           NewType(RecogMethodFamily,
+                   IsRecogMethod and IsAttributeStoringRep));
+
+# Functions and attributes for IsRecogMethod objects
+DeclareGlobalFunction("RecogMethod");
+DeclareGlobalFunction("CallRecogMethod");
+DeclareAttribute("FunctionOfRecogMethod", IsRecogMethod);
+DeclareAttribute("Stamp", IsRecogMethod);
+DeclareAttribute("Comment", IsRecogMethod);
```
fingolfin, line 39: It's a bit odd that one has the `...OfRecogMethod` suffix but the others have not... I realize `Function` isn't a great name, though :-).

But anyway, why do we even need `FunctionOfRecogMethod `? Only `CallRecogMethod` needs it, right? So that could just access the value directly (you can store things into objects in other ways than using an attribute!).

gap/base/methods.gi
```diff
@@ -0,0 +1,59 @@
...
+    SetStamp(r, stamp);
+    SetComment(r, comment);
+    SetFunctionOfRecogMethod(r, func);
+    return r;
+end);
+
+InstallGlobalFunction(CallRecogMethod,
+function(m, args)
+    return CallFuncList(FunctionOfRecogMethod(m), args);
```
fingolfin, line 58: If you change this to do something like `if IsRecord(m) then ... elif IsRecogMethod(m) then ... else Error() fi;` then you could immediately convert code to use it, right?

Alternatively, this could become an operation with several methods. Doesn't really matter, it's only temporary.

gap/base/methsel.gi
```diff
@@ -20,39 +20,46 @@
...
+    method := args[1];
+    if Length(args) = 1 then
+        rank := method.rank;
+        if not IsBound(method.comment) then
+            comment := "";
+        else
+            comment := method.comment;
+        fi;
+        method := RecogMethod(method.stamp, comment, method.method);
+        SetComment(method, comment);
```
fingolfin, line 36: This is redundant, isn't it?
```suggestion
```

gap/base/methsel.gi
```diff
@@ -81,33 +88,33 @@ InstallGlobalFunction( "CallMethods", function(db, tolerancelimit, methargs...)
...
 
             # skip methods which signalled a temporary failure at least
             # (tolerance + 1) times.
-            if IsBound(ms.failedMethods.(db[i].stamp)) and
-                ms.failedMethods.(db[i].stamp) > tolerance then
+            if IsBound(ms.failedMethods.(db[i].method.stamp)) and
+                ms.failedMethods.(db[i].method.stamp) > tolerance then
                 Info(InfoMethSel, 4, "Skipping rank ", db[i].rank,
-                     " method \"", db[i].stamp, "\".");
+                     " method \"", db[i].method.stamp, "\".");
```
fingolfin, line 103: If you add methods for `Stamp` and `Comment` using `InstallOtherMethod` which take an `IsRecord`, then perhaps you could immediately change this to
```suggestion
                     " method \"", Stamp(db[i].method), "\".");
```

fingolfin (commented):
Direction seems fine! Some comments on minor details, some are just opinions.

Don't forget to also adjust the HOWTO once everything is settled :-).

gap/base/methods.gd
ssiccha: The purpose of the `FunctionOfRecogMethod` is to provide a documented access to the underlying function. I think people should have acces to that and I didn't want to document access the function via e.g. `m!.func`.
But I think instead of the attribute I'll store the function in `m!.func` and provide an access function `Unpack`.

ssiccha: I've added a global function `UnpackRecogMethod`. I'm not married to the name though, so suggestions are always welcome. :)

gap/base/recognition.gi
```diff
@@ -451,7 +451,20 @@ InstallGlobalFunction( RecogniseGeneric,
...
+        tmp := [];
+        for hint in knowledge.hints do
+            if IsRecogMethod(hint.method) then
+                Add(tmp, rec(rank := hint.rank, method := hint.method));
+            else
+                Add(tmp,
+                    rec(rank := hint.rank,
+                        method := rec(method := hint.method, stamp := hint.stamp)));
+            fi;
+            allmethods := Concatenation(tmp, allmethods);
```
fingolfin, line 464: Huh, aren't you adding methods several times this way? You keep adding to tmp, then prepend it to allmethods -- again and again.

So either move this `Concatenation` out of the loop, or else just get rid of `tmp` and replace `Add(tmp` above by `Add(allmethods` (twice) -- the order should not matter as we sort by rank below anyway. 

Well, it might matter if several methods have identical rank, *and* if we were using `StableSortBy`.... I think having methods with equal rank is an error here, though, and perhaps we should add a corresponding assertion at the end. Something like this:
```gap
Assert(0, Length(Set(allmethods, m->m.rank)) = Length(allmethods)); # verify no rank occurs more than once
```

fingolfin (commented):
There is an error were something somewhere access `r.rank` which is not bound...

gap/base/methods.gi
```diff
@@ -0,0 +1,84 @@
...
+    s := "<RecogMethod \"";
+    Append(s, Stamp(m));
+    Append(s, "\": ");
+    Append(s, ViewString(UnpackRecogMethod(m)));
+    Append(s, ">");
+    return s;
+end);
+
+InstallGlobalFunction(RecogMethod,
+function(stamp, comment, func)
```
fingolfin, line 60: So there is no support for `validatesOrAlwaysValidInput` yet, which affects conversion of methods which set it.

That said, there isn't a *real* need to set `validatesOrAlwaysValidInput`, is there? It's not as if we are using it for anything. It's just a marker saying "someone went over this method and believes they have established that it is correctly validating everything". The one reason to have it as an explicitly variable is that it helps us to catch methods which don't have it yet?

gap/base/recognition.gi
fingolfin: added a fix

gap/projective/d247.gi
```diff
@@ -349,7 +349,9 @@ RECOG.SortOutReducibleSecondNormalSubgroup :=
 #! @BeginChunk D247
 #! TODO
 #! @EndChunk
-FindHomMethodsProjective.D247 := function(ri,G)
+FindHomMethodsProjective.D247 := RecogMethod("D247",
```
fingolfin, line 352: I suggest we add a helper `BindRecogMethod` to be used like this:
```suggestion
BindRecogMethod(FindHomMethodsProjective, "D247",
```

ssiccha:
Some commits are independent of this PR. I'll put them into a new PR.

gap/projective/d247.gi
fingolfin: I've now added `BindRecogMethod`

fingolfin:
@ssiccha I merged the independent commit (see PR #269) after a few extra tweaks, rebased this PR and cleaned it up.

I think now only `projective.gi` needs to be converted, then at least all calls to `AddMethods` only use RecogMethod objects. After these are done, AddMethods can be simplified again. Then there are likely in addition a few more methods only used in hints; once those are converted, the `allmethods` code can be simplified again, plus a bunch of other compatibility code be removed.

The `\=` method you added yesterday as a workaround for making `regen_doc.g` work can then also be removed again, and `regen_doc.g` be changed to instead just compare the stamp of methods to decide whether they are "equal".

gap/generic/KnownNilpotent.gi
```diff
@@ -104,14 +106,12 @@ FindHomMethodsGeneric.KnownNilpotent := function(ri,G)
   forfactor(ri).primes := primes{[1..cut]};
   forkernel(ri).primes := primes{[cut+1..Length(primes)]};
   Add(forfactor(ri).hints,
-      rec( method := FindHomMethodsGeneric.KnownNilpotent, rank := 4000,
-           stamp := "KnownNilpotent" ));
+      rec( method := FindHomMethodsGeneric.KnownNilpotent, rank := 4000 ) );
   Add(forkernel(ri).hints,
-      rec( method := FindHomMethodsGeneric.KnownNilpotent, rank := 4000,
-           stamp := "KnownNilpotent" ));
+      rec( method := FindHomMethodsGeneric.KnownNilpotent, rank := 4000 ) );
```
fingolfin, line 111: I wonder if we now also use `AddMethod` for these places? Like this:
```gap
      AddMethod(forkernel(ri).hints, FindHomMethodsGeneric.KnownNilpotent, 4000);
```

gap/base/methsel.gi
```diff
@@ -20,39 +20,45 @@
...
+    if not HasValidatesOrAlwaysValidInput(method) then
+        Info(InfoAddMethod, 1, "AddMethod, ", Stamp(method),
+             ": has no ValidatesOrAlwaysValidInput");
     fi;
     l := Length(methodDb);
-    pos := First([1 .. l], i -> methodDb[i].rank <= method.rank);
+    pos := First([1 .. l], i -> methodDb[i].rank <= rank);
     if pos = fail then
         pos := l + 1;
     fi;
```
fingolfin, line 54: I think this also works (but didn't test, beware of off-by-one errors etc.)
```suggestion
    pos := PositionSortedBy(methodDb, -rank, m -> -m.rank);
```

fingolfin, line 54: I think this also works (but didn't test, beware of off-by-one errors etc.)
```suggestion
    pos := PositionSortedBy(methodDb, -rank, m -> -m.rank);
```
and then the local variable `l` could be removed.

gap/base/methods.gi
ssiccha: At the moment, we can use it to check which methods don't have it yet, yes. I think in the long run it should be set to `true` for every method so we can remove it. Should we rather turn it into a source comment?

gap/base/methsel.gi
ssiccha: I'm going to install a method for `\<` for `RecogMethod`s, unless you object. Then we can use `PositionSorted` etc.

fingolfin: But why? We are comparing ranks, not recognition methods

ssiccha:
I think I'm finished with everything except for adjusting the method how to and double-checking whether I forgot to remove any hacks.

fingolfin:
@ssiccha regarding the commit "Make AddMethod compatible with GAP 4.10" -- Is there any value in keeping compatibility with GAP 4.10 at this point? We may want to use other improvements from 4.11 (or even 4.12, once it is out) anyway.

fingolfin (approved):

ssiccha:
@fingolfin I'd like to document `RecogMethod` using AutoDoc. However, with what I did in the last commit, the documentation of `RecogMethod` is put into a new chapter `methsel`. Do you know of a way to tell AutoDoc to include it into the chapter with GapDoc label `methsel`? Or is just not intended to start a chapter via GapDoc and then put stuff there via AutoDoc? I'm super confused. ^^

ssiccha:
I'm just going to use GAPDoc. I have no idea how to include it with AutoDoc doc.

fingolfin:
@ssicha Will you complete this? When?

Regarding AutoDoc and manuals: just write `@Chapter TITLE-OF-CHAPTER` before / at the start of your AutoDoc comment and you are good to go. No, there is no way to refer to a chapter by label, you need to repeat its title, but that's usually the same anyway.

ssiccha:
I've had a big teaching backlog to work through, but I'm done now.

I think I'm a bit dense when it comes to how to make GapDoc and AutoDoc work together. I've tried all possible combinations for `TITLE-OF-CHAPTER` I could think of:
`Method selection`, `Method_selection`, `Chapter_Method_selection`, ... The global function `RecogMethod` is always put into a completely new chapter. What am I doing wrong? For now I'm just going to use `@BeginChunk`.

ssiccha:
I've made some progress on the documentation. I'll improve it a little bit further, then this should be good.

fingolfin:
Sorry, I misunderstood: there is no way to write stuff in AutoDoc and "magically" insert it into an existing GAPDoc chapter. The only way to do that is indeed to use chunks. What I was talking about is only for adding content to a chapter "managed" by AutoDoc (which I should have realized is not the case here, but didn't).

ssiccha:
Rebased onto master. This is now ready for review @fingolfin.

gap/base/methods.gd
```diff
@@ -0,0 +1,130 @@
...
+BindGlobal("RecogMethodFamily", NewFamily("RecogMethodFamily", IsRecogMethod));
+BindGlobal("RecogMethodType",
+           NewType(RecogMethodFamily,
+                   IsRecogMethod and IsAttributeStoringRep));
+
+# <#GAPDoc Label="RecogMethod">
+# <ManSection>
+# <Func Name="RecogMethod" Arg="stamp, comment, func"/>
+# <Description>
+# Return a recognition method <C>method</C> in the filter
```
FriedrichRober, line 38: Maybe add a comment here, that `RecogMethod` should not be called directly but rather be created via `BindRecogMethod`.

ssiccha:
I've added a few tests just to get the coverage of `gap/base/methods.gi` to 100%.

gap/base/methods.gd
```diff
@@ -0,0 +1,130 @@
...
+# </Description>
+# </ManSection>
+# <#/GAPDoc>
+DeclareGlobalFunction("UnpackRecogMethod");
+
+# <#GAPDoc Label="CallRecogMethod">
+# <ManSection>
+# <Func Name="CallRecogMethod" Arg="m, args"/>
+# <Description>
+# Call <C>UnpackRecogMethod(m)</C> with arguments <A>args</A> and return the
```
FriedrichRober, line 101: Maybe add a comment here that `m(args)` does essentially the same.

gap/base/methods.gd
```diff
@@ -0,0 +1,130 @@
...
+# <#GAPDoc Label="UnpackRecogMethod">
+# <ManSection>
+# <Func Name="UnpackRecogMethod" Arg="method"/>
+# <Description>
+# Return the function underlying <A>method</A>.
+# The argument <A>m</A> must lie in <Ref Filt="IsRecogMethod"/>.
+# </Description>
+# </ManSection>
+# <#/GAPDoc>
+DeclareGlobalFunction("UnpackRecogMethod");
```
fingolfin, line 95: Is this really something we should expose to the user? What for? Once it's out, it's harder to take it back.

doc/methsel.xml
```diff
@@ -100,6 +103,31 @@ function:
...
+homomorphisms for group recognition.
+We collect them in these records such that we do not use up too many
+global variable names.
+
+<#Include Label="FindHomMethodsPerm">
+<#Include Label="FindHomMethodsMatrix">
+<#Include Label="FindHomMethodsProjective">
+<#Include Label="FindHomMethodsGeneric">
+
+This global record holds the functions for writing group
```
FriedrichRober, line 119: This line is confusing. To what global record are you referring? And why only matrix group recognition.

gap/base/methods.gi
```diff
@@ -0,0 +1,90 @@
...
+##  SPDX-License-Identifier: GPL-3.0-or-later
+##
+##
+##  Implementation of recog methods
+##
+#############################################################################
+
+InstallMethod(CallFuncList, [IsRecogMethod, IsList],
+{f, args} -> CallFuncList(f!.func, args));
+
```
fingolfin, line 21: Do we really want this? I think you are using `CallRecogMethod` explicitly after all, and I think this is indeed better -- it is easy to grep for `CallRecogMethod` and it is then possible to track what happens. But if one replace `CallRecogMethod(meth, val)` by `meth(val)` then this is rather intransparent, and may be very confusing, esp. to people who don't know that one can overload function call syntax in GAP (i.e., >99.9% of them)
```suggestion
```

gap/base/methsel.gi
```diff
@@ -20,40 +20,21 @@
...
-                      " method, rank, and stamp");
+InstallGlobalFunction("AddMethod", function(methodDb, method, rank)
+    local pos;
+    if not IsRecogMethod(method) then
+        ErrorNoReturn("<method> must be a RecogMethod");
     fi;
-    if not IsBound(method.comment) then
-        method.comment := "";
+    # This may be turned into an error in the future, see github issue #184.
+    if not HasValidatesOrAlwaysValidInput(method) then
```
fingolfin, line 29: I just merged another PR of your where you removed this code, I think?

gap/generic/KnownNilpotent.gi
```diff
@@ -76,7 +76,9 @@ end;
 #! p-group for some prime p. Make sure that the !.projective component is set
 #! correctly such that we can set the right Order method.
 #! @EndChunk
-FindHomMethodsGeneric.KnownNilpotent := function(ri,G)
+BindRecogMethod(FindHomMethodsGeneric, "KnownNilpotent",
+"TODO",
```
fingolfin, line 80: Perhaps
```suggestion
"method for nilpotent groups which are not p-groups",
```

gap/matrix.gi
```diff
@@ -125,7 +126,9 @@ end;
 #! @BeginChunk Scalar
 #! TODO
 #! @EndChunk
-FindHomMethodsMatrix.Scalar := function(ri, G)
+BindRecogMethod(FindHomMethodsMatrix, "Scalar",
+"Hint TODO",
```
fingolfin, line 130: Not for this PR, but: we should document these methods

gap/matrix.gi
```diff
@@ -214,7 +217,9 @@ end;
 #! will only be scalar matrices. This method recursively builds a balanced tree
 #! and does scalar recognition in each leaf.
 #! @EndChunk
-FindHomMethodsMatrix.BlockScalar := function(ri, G)
+BindRecogMethod(FindHomMethodsMatrix, "BlockScalar",
+"Hint TODO",
```
fingolfin, line 221: Not for this PR, but: we should document these methods

gap/matrix.gi
```diff
@@ -479,7 +478,9 @@ end;
 #! recognition info record has the <C>!.projective</C> component bound
 #! to <K>true</K> and this information is passed down to image and kernel.
 #! @EndChunk
-FindHomMethodsMatrix.BlockLowerTriangular := function(ri,G)
+BindRecogMethod(FindHomMethodsMatrix, "BlockLowerTriangular",
+"Hint TODO",
```
fingolfin, line 482:
```suggestion
"for a group generated by block lower triangular matrices",
```

gap/matrix.gi
```diff
@@ -525,18 +523,17 @@ end;
 #! recognition info record has the <C>!.projective</C> component bound
 #! to <K>true</K> and this information is passed down to image and kernel.
 #! @EndChunk
-FindHomMethodsMatrix.BlockDiagonal := function(ri,G)
+BindRecogMethod(FindHomMethodsMatrix, "BlockDiagonal",
+"Hint TODO",
```
fingolfin, line 527:
```suggestion
"for groups generated by block diagonal matrices",
```

fingolfin (approved):

gap/base/methods.gd
FriedrichRober: After reading the review of Max, probably it would be better to drop `m(args)` entirely.

gap/base/methods.gi
FriedrichRober: I agree, we should probably drop `meth(val)`.

gap/base/methods.gd
ssiccha: I've removed the documentation and the `Declare` statement.

gap/base/methods.gi
ssiccha: Makes sense. I've removed this.

gap/base/methsel.gi
ssiccha: I rebased it. Now it's gone here too.

gap/base/methods.gd
ssiccha: I've removed the `m(args)` functionality as @fingolfin suggested.

doc/methsel.xml
ssiccha: Ooh, yes. Thanks for spotting that.

fingolfin:
@ssiccha hmmm, now all tests are failing

ssiccha:
Jepp. I'm looking into it. I suspect its because I rebased everything on top of master.

ssiccha:
I added, but didn't document, the ability to call a recog method via `m(args)`. It's used in several places of recog. I also adjusted the SnAnUnknownDegree code to this PR. Tests pass on my machine now. Let's see whether they also pass on GitHub. Then I'll merge this.

PR #264, opened 2021-04-08
Separate methods and ranks

ssiccha:
This is preparation for #263.


ssiccha:
This is very hacky. But I'm currently replacing the method records by `IsRecogMethod` objects in #263 anyways and so I don't want to change all `AddMethods` calls twice.

ssiccha:
Actually, I think I'll just leave this as a part of #263.

PR #265, opened 2021-04-09
Implement extensions for `SnAnUnknownDegree` by Conder and Holt

### LINKED ISSUE #348: SnAn naming algorithm
Related to #265. We included code from magma to guess the degree of a potentially alternating or symmetric group. However, this code tests a lot of element orders and creates a lot of overhead. In particular, it does not recognise for many groups that they cannot be Sn or An. @aniemeyer suggested that we could improve on this method by first determining a lower bound for the degree `n` (like in magma) but with far fewer tries, then taking this initial guess and check if elements of order close to `n` (n-cycles) can be found in e.g. `2n` tries. We might need to revise `n` upwards after these trials. It is also worth looking into `FindHomMethodsProjective.AltSymBBByDegree` if something in this code is already attempting to guess the degree.

ssiccha:
A PR based on #176.

We are going to implement the extensions by Conder and Derek Holt's
GuessAltSymDegree.


ssiccha:
Rebased the last commit onto the current version of #176.

FriedrichRober:
Link to our [Hackmd](https://hackmd.io/Ts8uHyVCRICFEqxBy_5Gtw?view)

ssiccha:
We should be almost done. We think there are two TODOs left: clean up the tests and figure out what to do with small classical groups like: `Omega(-1,4,3)`. It has element orders <= 5, thus slips through our heuristics to weed out groups which are not isomorphic to An or Sn. Hence the SnAnUnknownDegree algorithm takes forever until it realizes, that the group in question is not an An or Sn. We think we should be able to alleviate this by simply saying that we're NeverApplicable if the vector space our group acts has less than e.g. 100 elements.

fingolfin:
Awesome, thank you!

fingolfin:
@FriedrichRober do you think you can finish this on your own, now that @ssiccha is gone? Perhaps I can help a bit?

FriedrichRober:
Yes, I think I can finish this on my own. The code is finished (except for handling small classical groups). What is missing now is to finish the documentation (comments) and looking at the runtime of the test suite.

FriedrichRober:
@fingolfin, ich habe nun die Änderungen implementiert, wie wir es besprochen haben. Wir geben früher auf, und speichern uns die Iteratoren im recog node ab, um sie später weiter zu verwenden. Trotzdem dauert die Test-Suite auf meinem Rechner 3-4 Minuten länger als auf dem master. Meiner Meinung nach ist dieser Zustand inakzeptabel, um das einfach so zu mergen :(

Ich muss dann nochmal genau schauen, an welchen Stellen wir zu viel Zeit verschwenden, und ob sich da was machen lässt.

FriedrichRober:
We can see which tests fail after everything has run. I think I remember that some small classical groups were recognised now as alternating or symmetric and therefore destroy the output of some tests.

FriedrichRober:
Exactly, for example GL(4,2) is now recognized as A8, which is correct but not what the test file expects

FriedrichRober:
@fingolfin, the pull request is ready now. The test suite on my laptop was as quick as on master, however some unrelated errors occures. Two where the stamp is not what was expected in the test file. One where the group was wrongly recognized which would require further investigation. I assume that this is a random error which simply occured from inserting the recognition method into the database.

fingolfin:
Fantastic, @FriedrichRober, thank you so much!

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -37,12 +111,35 @@ RECOG.ThreeCycleCandidatesConstants := function(eps, N)
...
+    if not isone(ri)(c ^ 3) then
+        return false;
+    fi;
+    for k in [1 .. logInt2N + 1] do
+        r := R[k];
+        # c * c ^ r is a product of two three-cycles, so it should have order
+        # 1, 2, 3 or 5.
+        y := c * c ^ r;
+        yTo5 := y ^ 5;
+        if not isone(ri)(yTo5) and not isone(ri)(yTo5 * y) then
```
fingolfin: Computing $y^5 = $(y^2)^2\cdot y$ require three multiplications, and then $y^5\cdot y$ a fourth one. So either 3 or 4 multiplications

Wouldn't it then be even better to do something like this, which requires either 1, 2 or 3 multiplications ?
```suggestion
        y2 := y^2;
        if not isone(ri)(y2) then
          y3 := y2 * y;
          if not isone(ri)(y3) and not isone(ri)(y2 * y3) then
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -700,49 +1009,52 @@ RECOG.ConstructSnAnIsomorphism := function(ri, n, stdGensAnWithMemory)
...
-# probability at least 1 - <A>eps</A> the degree <A>n</A> and an
-# isomorphism from <A>G</A> to An or Sn.
-RECOG.RecogniseSnAn := function(ri, eps, N)
-    local T, foundPreImagesOfStdGens, constants, iterator, c, tmp, recogData, i;
-    T := Int(Ceil(Log2(1 / Float(eps))));
-    foundPreImagesOfStdGens := false;
-    constants := RECOG.ThreeCycleCandidatesConstants(1. / 4., N);
-    for i in [1 .. T] do
-        iterator := RECOG.ThreeCycleCandidatesIterator(ri, constants);
+# This is the main method used by RECOG.RecogniseSnAn and RECOG.RecogniseSnAnLazy.
```
fingolfin: As far as I can tell, `RECOG.RecogniseSnAn` is not used anymore. Can we just remove it then, and perhaps rename `RECOG.RecogniseSnAnLazy` to `RECOG.RecogniseSnAn` ?

fingolfin (commented):
Thank you! I've pushed a bunch of changes (see commit log). I'll try to understand the errors better tomorrow

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: yes, your suggestion is definitely better

FriedrichRober:
> Thank you! I've pushed a bunch of changes (see commit log). I'll try to understand the errors better tomorrow

The rest of the changes look good to me

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -37,12 +111,35 @@ RECOG.ThreeCycleCandidatesConstants := function(eps, N)
...
+        if not isone(ri)(yTo5) and not isone(ri)(yTo5 * y) then
+            return false;
+        fi;
+    od;
+    return true;
+end;
+
 # ri : recognition node with group G
-# constants : a record with components M, B, T, C, and logInt2N
+# constants : a record with components M, B, T, C, K, L, and logInt2N
```
TWiedemann:
```suggestion
# constants : a record with components M, B, T, C, K, L, and logInt2N
# whose meaning is documented at the beginning of this file
```
Add ref to documentation.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -61,103 +165,180 @@ RECOG.ThreeCycleCandidatesIterator := function(ri, constants)
...
+            if curInvolutionPos > Length(involutions) then
+                r := RandomElm(ri, "SnAnUnknownDegree", true)!.el;
+                # In the paper, we have t = r ^ M.
+                # Invariant: tPower = (r ^ M) ^ (2 ^ a)
+                tPower := r ^ M;
+                # We make a small improvement to the version described in
+                # <Cite Key="JLNP13"/>. The order of r ^ M is a 2-power.
+                # It can be at most 2 ^ logInt2N. Thus, if we find an r such that
+                # (r ^ M) ^ (2 ^ logInt2N) is non-trivial, then we can return
+                # NeverApplicable.
```
TWiedemann:
```suggestion
                # We make a small improvement to the version described in
                # <Cite Key="JLNP13"/>. The order of r ^ M is a 2-power.
                # It can be at most 2 ^ logInt2N. Thus, if we find an r such that
                # (r ^ M) ^ (2 ^ logInt2N) is non-trivial, then we can return
                # NeverApplicable. I.e. we need one iteration less than in [JLNP13],
                # which also computes (r ^ M) ^ (2 ^ (longInt2N + 1)).
```
Make the difference between the code and JLNP13 more explicit (hoping that I understood this correctly, I am not sure).

TWiedemann:
@FriedrichRober I went through `ThreeCycleCandidatesIterator` and it looks good to me! The only other comment I have is that I think whenever you write "commutating", you actually mean "non-commutating": The algorithm looks for conjugates of the involution `t` that do not commute with `t`, and counts how many it has found already. Hence I think `nrCommutatingConjugates` should be appropriately renamed, and some comments adapted.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -61,103 +165,180 @@ RECOG.ThreeCycleCandidatesIterator := function(ri, constants)
...
+
+    R := List([1 .. logInt2N + 1], k -> StripMemory(RandomElm(ri, "SnAnUnknownDegree", true)!.el));
 
     # Counters
-    # Counts the constructed involutions t_i in steps 2 & 3.
-    nrInvolutions := 0;
+    curInvolutionPos := 1;
+    Ki := Minimum(K, C);
+    Li := Minimum(L, B);
     # Counts the elements c in step 4 that we use to conjugate the current
```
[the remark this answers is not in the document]
aniemeyer: It is actually 'commuting' and 'non-commuting' instead of 'commutating' and 'non-commutating"

TWiedemann:
The implementation of `RECOG.FindAnElementMappingIToJ` and `RECOG.FindImageSnAnSmallDegree` looks flawless to me.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
...
-# TODO: Image Computation requires n >= 11.
+# - slpToStdGens: an SLP from the generators of G to stdGens.
 RECOG.ConstructSnAnIsomorphism := function(ri, n, stdGensAnWithMemory)
-    local stdGensAn, xis, gImage, foundOddPermutation, slp, eval,
+    local stdGensAn, filter, filterPart, gImage, foundOddPermutation, slp, eval,
         hWithMemory, bWithMemory, stdGensSnWithMemory, b, h, g;
     stdGensAn := StripMemory(stdGensAnWithMemory);
-    xis := RECOG.ConstructXiAn(n, stdGensAn[1], stdGensAn[2]);
+    # Construct filter. In <Cite Key="C12"/> filter is called E.
+    # In <Cite Key="JLNP13"/> filter is called `xi`.
```
TWiedemann:
```suggestion
    # In <Cite Key="JLNP13"/> filter is called `x_i`.
```
Is this correct? I initially thought the comment referred to the Greek letter \xi.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -752,41 +1069,193 @@ RECOG.RecogniseSnAn := function(ri, eps, N)
...
+    # mindego and mindege will be respectively the smallest possible
+    # degrees of symmetric groups that contain the elements of odd and
+    # even orders, in the random sample.
+    # If mindego > mindege we assume the group is alternating, otherwise
+    # that it is symmetric.
+
+    G := Grp(ri);
+    if  (IsPermGroup(G) and NrMovedPoints(G) <= 6)
+                or (IsMatrixGroup(G) and DimensionOfMatrixGroup(G) < 3) then
+        Print("GuessAltsymDegree works only for degree > 6\n");
```
TWiedemann:
```suggestion
        Info(InfoRecog, 2, "GuessSnAnDegree works only for permutation degree > 6 or matrix dimension > 2");
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
     fi;
 end;
 
-# This function is an excerpt of the function RECOG.RecogniseSnAn in gap/SnAnBB.gi
+# The following two functions RECOG.FindAnElementMappingIToJ and
+# RECOG.FindImageSnAnSmallDegree are used for small degrees, 5 <= n < 11, to
+# compute a monomorphism into Sn based on Jonathan Conder's thesis <Cite
+# Key="C12"/>.
+#
```
fingolfin, line 848:
```suggestion
#
# Under that monomorphism s is mapped to a long cycle, t to a three cycle,
# and the list e (E in Conder's Thesis) to [(1,2,3), (1,2,4), (1,2,5)].
#
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
...
+        return s^(j-i);
+    fi;
+end;
+
+# In Conder's Thesis: Algorithm 10, ElementToSmallDegreePermutation,
+# correctness see Theorem 3.5.2.
+# Returns the image of g under the monomorphism Grp(ri) -> S_n given by s and
+# t, for n >= 5.
+# Under that monomorphism s is mapped to a long cycle, t to a three cycle,
+# and the list e (E in Conder's Thesis) to [(1,2,3), (1,2,4), (1,2,5)].
```
fingolfin:
```suggestion
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
     fi;
 end;
 
-# This function is an excerpt of the function RECOG.RecogniseSnAn in gap/SnAnBB.gi
+# The following two functions RECOG.FindAnElementMappingIToJ and
+# RECOG.FindImageSnAnSmallDegree are used for small degrees, 5 <= n < 11, to
+# compute a monomorphism into Sn based on Jonathan Conder's thesis <Cite
+# Key="C12"/>.
```
fingolfin:
```suggestion
# Key="C12"/>, Definition 3.2.2.
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
...
+#     RECOG.FindImageSn, RECOG.FindImageAn, and RECOG.FindImageSnAnSmallDegree
+#     to compute the isomorphism.
 #   - n is the degree of the group.
-# - slpToStdGens: an SLP to stdGens.
-#
-# TODO: Image Computation requires n >= 11.
+# - slpToStdGens: an SLP from the generators of G to stdGens.
 RECOG.ConstructSnAnIsomorphism := function(ri, n, stdGensAnWithMemory)
-    local stdGensAn, xis, gImage, foundOddPermutation, slp, eval,
+    local stdGensAn, filter, filterPart, gImage, foundOddPermutation, slp, eval,
```
fingolfin: Why is this called `filter`? It is neither a filter in GAP's sense, nor in the mathematical sense (think "ultrafilter") nor do any of the sources use this term. I find it rather confusing.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -641,32 +822,165 @@ RECOG.StandardGenerators := function(ri, g, c, k, eps, N)
...
         hWithMemory, bWithMemory, stdGensSnWithMemory, b, h, g;
     stdGensAn := StripMemory(stdGensAnWithMemory);
-    xis := RECOG.ConstructXiAn(n, stdGensAn[1], stdGensAn[2]);
+    # Construct filter. In <Cite Key="C12"/> filter is called E.
+    # In <Cite Key="JLNP13"/> filter is called `xi`.
```
fingolfin: It would be better if we had more concrete refs (so to specific definitions/lemmas); as it is we could not quite tell which "xi" or "Xi" or "X_i" or "x_i" are meant exactly.

I also note that `RECOG.ConstructXiAn` comes from the older code written for the paper by Beals et all (`BLN+03` here?), so perhaps that is also a reference?

(Of course the fact that `RECOG.ConstructXiAn` in turn is undocumented doesn't help either...)

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -752,41 +1069,192 @@ RECOG.RecogniseSnAn := function(ri, eps, N)
...
-    eps := 1 / 10^2;
+# This method is an implementation of <Cite Key="JLNP13"/>.
+# From <Cite Key="JLNP13" Where="Theorem 1.1"/>:
+# RECOG.RecogniseSnAn is a one-sided Monte-Carlo algorithm with the following
+# properties. It takes as input a black-box group <A>G</A>, a natural number
+# <A>N</A> and a real number <A>eps</A> with 0 < <A>eps</A> < 1. If <A>G</A> is
+# isomorphic to An or Sn for some 9 <= <A>n</A> <= <A>N</A>, it returns with
+# probability at least 1 - <A>eps</A> the degree <A>n</A> and an
+# isomorphism from <A>G</A> to An or Sn.
+RECOG.RecogniseSnAn := function(ri, eps, N)
```
fingolfin: As discussed above, should perhaps be renamed, and a comment added that we don't use it and instead use the "lazy" version.

If we do want to keep this unused code, then I insist that we at least add some test cases for it, otherwise it will just bitrot and help nobody.

As to renaming, above `RECOG.RecogniseSnAnExact` was suggested as a new name, but this is not clear to me. "Exact" in which sense? Perhaps "exactly as in the paper" -- but there are  other meaning of "Exact" and I don't think I'd guess that particular one.

If we stick with the "lazy" idea, perhaps `RECOG.RecogniseSnAnEager` would be better? Or "...Greedy" or ...

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -752,41 +1069,192 @@ RECOG.RecogniseSnAn := function(ri, eps, N)
...
+    local G, r, options, mintries, maxtries, fac, mindego, mindege, ct, cto, cte, proc, g, o, mindeg, o_fact, mindegforg;
+    # mindego and mindege will be respectively the smallest possible
+    # degrees of symmetric groups that contain the elements of odd and
+    # even orders, in the random sample.
+    # If mindego > mindege we assume the group is alternating, otherwise
+    # that it is symmetric.
+
+    G := Grp(ri);
+    if  (IsPermGroup(G) and NrMovedPoints(G) <= 6)
+                or (IsMatrixGroup(G) and DimensionOfMatrixGroup(G) < 3) then
```
fingolfin:
```suggestion
                or (IsMatrixGroup(G) and DimensionOfMatrixGroup(G) <= 2) then
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -826,8 +1294,96 @@ function(ri, G)
...
+# computations. It is the main function of SnAnUnknownDegree.
+RECOG.RecogniseSnAnLazy := function(ri)
+    local a, cache, N, tmp;
+    RECOG.SnAnCacheUpperBoundForDegree(ri);
+    cache := RECOG.SnAnGetCache(ri);
+    N := cache.N;
+    if N = TemporaryFailure then
+        RECOG.SnAnResetCache(ri);
+    fi;
+    if not IsInt(N) then
```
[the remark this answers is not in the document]
fingolfin:
```suggestion
    if N = TemporaryFailure or N = NeverApplicable then
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -826,8 +1294,96 @@ function(ri, G)
...
+#! @EndChunk
+BindRecogMethod(FindHomMethodsGeneric, "SnAnUnknownDegree",
+"method groups isomorphic to Sn or An with n >= 9",
+function(ri, G)
+    local recogData, isoData, degree, swapSLP, t;
     # Try to find an isomorphism
-    recogData := RECOG.RecogniseSnAn(ri, eps, N);
+    recogData := RECOG.RecogniseSnAnLazy(ri);
     # RECOG.RecogniseSnAn returned NeverApplicable or TemporaryFailure
     if not IsRecord(recogData) then
```
fingolfin:
```suggestion
    if recogData = NeverApplicable or recogData = TemporaryFailure then
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,7 +17,81 @@
 ##  <Cite Key="JLNP13"/>.
 ##
 #############################################################################
-#
+
+##  General Strategy
+##  ----------------
+##
+##  Here we describe the general strategy that is used to recognise if a group
+##  is ismorphic to `An` or `Sn` for `n >= 7`.
```
fingolfin: But below when defining `SnAnUnknownDegree` (around line 1355ff), it is stated that this method is for $n\geq 9$. Some parts of the code suggest that certain operations require $n\geq 9$ and others $n\geq 7$, which might explain the difference (and of course it would mean overall we indeed need $n\geq 9$).

And then there is the code in `RECOG.SnAnUpperBoundForDegree` which kinda seems to suggest we might have to assume $n\geq 10$, at least for non-projective matrix groups?!

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,7 +17,81 @@
...
+##
+##  We have to be careful what we do for small degrees.
+##  If we pass an `A5`, `S5`, `A6`, `S6` into `SnAnUnknownDegree`, then it will not recognize it.
+##  If the input acts on a space with a large dimension, then this can take forever.
```
fingolfin: Is that last bit still the case? If so, a concrete example would be good so we can better guard against it (i.e., an example could be used in a test that verifies that whatever "guards" we use are effective)

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,7 +17,81 @@
...
+##
+##  - We assume that our input group `G` is a projective irreducible matrix group.
```
fingolfin: Ah, OK. That might partially address my concern from above about `RECOG.SnAnUpperBoundForDegree` requiring $n\geq 10$ in the non-projective case.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,7 +17,81 @@
...
+##  - Deduce an upper bound `N` for the degree of `An`, `Sn` by using bounds provided in [KL90].
+##  - Look at some orders and deduce a lower bound `M` for the degree.
+##      - If `M > N`, then we excluded `An`, `Sn`.
+##      - If `M <= 6`, then we return `TemporaryFailure`.
+##      - TODO: If `N > 20`, use Magma Code `GuessSnAnDegree` to guess the degree by element orders.
+##      - TODO: If `GuessSnAnDegree` was used, then we check if we can find
+##        an element of order in the interval `[N - m, N + m]`.
```
fingolfin: A word or number seems to be missing after "an element of order" ?

tst/working/slow/GenericSnAnUnknownDegree.tst
```diff
@@ -1,310 +0,0 @@
...
-# An
-gap> for d in [11 .. 30] do
-> sets := Combinations([1 .. d], 2);;
-> SdOn2Sets := Action(AlternatingGroup(d), sets, OnSets);;
-> ri := RecogNode(SdOn2Sets);;
-> success := FindHomMethodsGeneric.SnAnUnknownDegree(ri);
-> if not success or not Size(ri) = Factorial(d)/2 then
->   Print("wrong result! degree ", d, "\n");
-> fi;
-> od;
```
fingolfin, line 291: Removing this will result in quite a loss of coverage. Why do we do it?

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: Let me check this quickly so you don’t end up on a wild goose chase.

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: Yes, this is indeed from BLN+03. It is introduced in prose, so has no reference, and then is used for the first time in Lemma 5.3. If n>=11, then this filter can be used to find the image under the homomorphism. For degree <11, we had to use Conder's algorithm for the image computation, which works for degree>=5.

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: If I remember correctly, we used the word "filter" because the xis act like a sieve or filtering mechanism on the search space for finding the image points. But any renaming would be fine here :)

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: I think we can recognise groups with degree n>=7, and that groups with degree 5 or 6 cannot be recognised.

Details:
- Lower Bounds need n >= 9:
This is just to satisfy the error-probabilities, and has no impact on the correctness of the algorithm.
- Bolstering Elements are only defined for n >= 7:
This is the crucial point, which I think would lead to the algorithm returning fail for smaller degrees.
- In SnAnUpperBoundForDegree:
   - For the projective groups, the relevant bounds are only defined for n≥9. To handle all cases, we set N:=max(8,d+2). When the actual degree is at least 9, we use the corresponding bound; otherwise, we assume the worst case, namely degree at most 8. Hence, the argument remains valid for all degrees.
   - Same holds for non-projective matrix groups with n>=10.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: "element of order in the interval [N - m, N + m]" sounds fine to me? As in, "element whose order is at least N-M and at most N+m"?

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: I think it will not anymore run forever, since I adjusted the code to return fail more quickly. But I agree that a test case would be nice to confirm this, see my comment about bolstering elements only being defined for n>=7.

FriedrichRober: Anyway, I also think that the SnAnUnknownDegree code assumes that the input is irreducible (because it runs after FindHomMethodsProjective.NotAbsolutelyIrred in the recog framework, etc.), so these examples that Sergio and I thought of as problematic should not be passed as input to SnAnUnknownDegree at this stage.

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: I think “element whose order is in [N-m, N+m]” is more clear

SoongNoonien:
```suggestion
##        an element whose order is in the interval `[N - m, N + m]`.
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,11 +17,89 @@
...
+##  If the input acts on a space with a large dimension, then this can take forever.
+##
+##  - We assume that our input group `G` is a projective irreducible matrix group.
+##  - Deduce an upper bound `N` for the degree of `An`, `Sn` by using bounds provided in [KL90].
+##  - Look at some orders and deduce a lower bound `M` for the degree.
+##      - If `M > N`, then we excluded `An`, `Sn`.
+##      - If `M <= 6`, then we return `TemporaryFailure`.
+##      - TODO: If `N > 20`, use Magma Code `GuessSnAnDegree` to guess the 
+##        degree by element orders. (In Issue 348, there is code for a GAP function
+##        `GuessSnAnDegree` which can be used a starting point.)
```
SoongNoonien:
```suggestion
##        `GuessSnAnDegree` which can be used as a starting point.)
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -61,103 +169,180 @@ RECOG.ThreeCycleCandidatesIterator := function(ri, constants)
...
+    Li := Minimum(L, B);
+    # Entry i of this list counts the elements c in step 4 that we use to conjugate the current
+    # involution t_i.
+    nrTriedConjugates := [];
+# Entry i of this list counts the size of the set Gamma_i in step 4
+# for the current involution t_i
+    nrCommutatingConjugates := [];
+    # Entry i of this list counts the actual number of three cycle candidates considered
+    # which are filtered from the Gamma_i via heuristic order tests.
+    nrThreeCycleCandidates := [];
```
fingolfin, line 220:
```suggestion

    # Entry i of this list counts the elements c in step 4 that we use to conjugate the current
    # involution t_i.
    nrTriedConjugates := [];

    # Entry i of this list counts the size of the set Gamma_i in step 4
    # for the current involution t_i
    nrCommutatingConjugates := [];

    # Entry i of this list counts the actual number of three cycle candidates considered
    # which are filtered from the Gamma_i via heuristic order tests.
    nrThreeCycleCandidates := [];
```

gap/generic/SnAnUnknownDegree.gi
SoongNoonien: This is discussed in another thread below. I'll resolve this thread here to reduce at least my confusion a bit.

tst/working/slow/GenericSnAnUnknownDegree.tst
SoongNoonien: I've restored them now.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: I have added a commit 7449f57 which renames "filter" by "sieve", which resolves this issue.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: Commit 7449f57 appends "Section 3.4" to the reference C12, making it more precise.
I was not able to find `xi` in JLNP13.
Commit 464b160 adds a reference to "BLN+03, Lemma 5.3 and the paragraphs preceding it" to the documentation of RECOG.ConstructXiAn, which was the best I could figure out.

gap/generic/SnAnUnknownDegree.gi
fingolfin: I guess if the code requires irreducibility or even absolute irreducibility, then it should check `RECOG.IsIrreducible(ri)` resp. `RECOG.IsAbsolutelyIrreducible(ri)` and return `NeverApplicable` if the result is `false`

fingolfin (commented):
we agreed during the last meeting that this is basically ready, once CI passes. Which it not mostly does, one error remains, only in the test against GAP devel:

```
########> Diff in /tmp/gaproot/pkg/recog/tst/working/slow/bugfix.tst:25
# Input is:
for i in [1..50] do
    ri := RECOG.TestGroup(GL(8,27), false, Size(GL(8,27)));
od;
# Expected output:
# But found:
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `Length' on 1 arguments
The 1st argument is 'fail' which might point to an earlier problem

```
From a quick local test, it seems as if there is an error deep inside the $SL_n$ recognition, where it unexpectedly gets a `fail` in a place this is not expected. I have yet to dig into it. But it seems more like an issue that already exists in `master`, i.e. is not caused by this PR, but surfacing because it... But I might also be wrong on that, we'll have to see.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: The comment in line 31 of SnAnUnknownDegree.gi indeed claims "We assume that our input group `G` is a projective irreducible matrix group.". By what @FriedrichRober says about where this code runs in the recog framework, it seems that we *can* assume irreducibility, but it is not clear to me that the code actually requires irreducibility. 

The claim (in line 31) that projectivity of G is assumed also confuses me. The function SnAnUpperBoundForDegree has a case distinction that covers projective and non-projective matrix groups separately (in particular, non-projective matrix groups are covered, at least by this function), and projectivity does not seem to pop up anywhere else. Maybe line 31 should say "possibly projective"?

The reference [JLNP13], which describes the vanilla algorithm, is about black box groups, and hence does not reference irreducibility or projectiveness.

FriedrichRober: Yes, we do not require any additional assumptions to run the algorithm per se.

The upper bounds we use for the permutation degree N are obtained by looking at the matrix degree of the smallest (irreducible) (projective) representation that can occur for any given permutation degree n. But to apply the bounds we do not require the input group to be irreducible.

The reason we want to assume irreducibility is
1) it is (probably?) cheaper to run meat-axe than this
2) edge cases of small degree representations (n=5,6) acting on large vector spaces (reducible reps) might waste too much time without being able to produce an isomorphism (I did not test if the fixes I applied in the lazy variant resolved the issue with this)

Feel free to ping me if my explanations are unclear.

TWiedemann: @FriedrichRober Thanks, that helps!

@fingolfin I conclude from Friedrich's explanation that we should indeed return `NeverApplicable` if the group is not irreducible, and give the reasons that Friedrich stated above in the documentation.

TWiedemann: Commit 737d701 adds the `RECOG.IsIrreducible(ri)` check and adds Friedrich's explanation to the comments.

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -17,11 +17,91 @@
...
+##  ----------------
+##
+##  Here we describe the general strategy that is used to recognise if a group
+##  is isomorphic to `An` or `Sn` for `n >= 7`.
+##
+##  We have to be careful what we do for small degrees.
+##  If we pass an `A5`, `S5`, `A6`, `S6` into `SnAnUnknownDegree`, then it will not recognize it.
+##  If the input acts on a space with a large dimension, then this can take forever.
+##
+##  - We assume that our input group `G` is a (possibly projective) irreducible matrix group.
```
SoongNoonien: Do we actually assume that group is a matrix group? Telling from the code it seems like we also give helpful results if the group is a permutation group.

fingolfin: It better does, for we do this:
```gap
AddMethod(FindHomDbPerm, FindHomMethodsGeneric.SnAnUnknownDegree, 58);;
```

gap/generic/SnAnUnknownDegree.gi
```diff
@@ -826,10 +1183,83 @@ function(ri)
...
+#! If the input group is isomorphic to a symmetric or alternating group of
+#! degrees 5 or 6, then this method might not exit quickly.
+#!
+#! @EndChunk
+BindRecogMethod("FindHomMethodsGeneric", "SnAnUnknownDegree",
+"method for groups isomorphic to Sn or An with n >= 7",
+function(ri)
+    local recogData, isoData, degree, swapSLP, t;
+    # If the representation is a reducible matrix group, we should try something else
+    if IsMatrixGroup(Grp(ri)) and not RECOG.IsIrreducible(ri) then
```
TWiedemann, line 1262: @SoongNoonien and I were unsure whether the irreducibility check should be performed here or in `SnAnUpperBoundForDegree`. On the one hand, `SnAnUpperBoundForDegree` already has a case distinction for `IsPermGroup`/`IsMatrixGroup`, so the check would fit there. On the other hand, since the irreducibility of the matrix group is an assumption on the input, one might argue that it should be checked in the beginning.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: @FriedrichRober Can the input `G` be an arbitrary permutation group, or do we need to make further assumptions? The comments in `SnAnUpperBoundForDegree` say "We assume that G is primitive and not a giant in natural representation.", but maybe this is only to establish the bounds, and not an assumption on the input?

fingolfin: If the bounds assume primitivity, then we should only use them if primitivity is a given, right? So either refuse to run on imprimitive inputs, or use different bounds.

Note that there are already reduction methods in place which are supposed to deal with non-transitive and with imprimitive permutation groups, which ought to be called *before* the Sn/An methods. Thus it is fine to require primitivity; but this still should be checked for, to avoid potential bugs (e.g. due to some of these earlier test having a `TemporaryFailure`)

gap/generic/SnAnUnknownDegree.gi
fingolfin, line 1154: I realize this code was here before this PR, but while reading through things to figure out if and where to add a primitivity check, I stumbled over this: this comment reads to me as if `N` is a *lower* bound for `n` here? Maybe I am misunderstanding something?

fingolfin: I find this comment (which predates this PR) confusing. First we assume `N >= 9`, but then somehow arrive at a value `N < 9`... but we don't say this is a contradiction, but rather conclude that `N < 9`, which sounds tautological. 

I think something like this is meant:


```suggestion
# If n >= 9, then the smallest irreducible projective An-module has
# dimension n-2, see [KL90], Proposition 5.3.7.
# Turning things around, this means d+2 is an upper bound for the degree,
# at least if d+2 >= 9. Otherwise n < 9 and so then 9 is an upper bound.
```

Hrm... I am still not 100% satisfied with this, gotta think a bit more.

Of course the "else" branch has a similar comment with similar issues.

gap/generic/SnAnUnknownDegree.gi
fingolfin: I think it is good to do here; we should likewise check for primitivity here, in case of a perm group

gap/generic/SnAnUnknownDegree.gi
TWiedemann: I agree that
1. what you wrote is what is meant in the original comment (at least that is the way I understood it),
2. your comment is more clear.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: I was thinking that maybe the situation is similar to the one for matrix groups: We first obtain a bound for primitive permutation groups by finding the smallest degree of a primitive action of $S_n$ and $A_n$, and then (by some arguments) the same bound also holds for non-primitive groups. In the matrix group case, this works (with "primitive" replaced by "irreducible"): If every irreducible representation of a group $G$ has dimension at least $d$, then every representation of $G$ has dimension at least $d$ (because every representation contains an irreducible representation). Does a similar argument work for primitive action, i.e. every action of degree $d$ induces a primitive action of degree at most $d$? I think not, but I am not sure. If not, then I think we should indeed test primitivity of the input.

In any case, if we are not sure we can just require primitivity and it should not cause a problem by what you say about other reduction methods being applied earlier.

gap/generic/SnAnUnknownDegree.gi
TWiedemann: I think what is meant in the comment is actually `n <= 1/2 + Sqrt(1/4 + 2*k)`. The sentence "Thus its degree is at least n * (n-1) / 2" means that
$k \ge n * (n-1)/2$.
Hence
$2k \ge n(n-1) \ge (n-1)^2$.
Hence
$n-1 \le \sqrt{2k}$.
Hence
$n \le \sqrt{2k}+1$.
This is not exactly the bound written in the comment, but somewhat close (and I do not understand where exactly the bound in the comment comes from).

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: The bounds are definitely assuming that it is not the natural action on n points, so the giant method should run before this method (it is also cheaper).

We also assume primitivity for the bounds. A similar argument as for irreducible does not apply here. A faithful imprimitive action does not induce a faithful primitive action of smaller degree, the faithful part is what is important here.

Take as an example S_2 wr S_2. The imprimitive action on 2 blocks of size 2 is on 4 points. However, the group does not even admit any faithful primitive action, since it does not have any corefree maximal subgroups.

Other similar example, S_5 wr S_2 has imprimitive action on 10 points, but smallest primitive action on 25 points (e.g. product action), since the largest maximal corefree subgroup has index 25.

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: If k is the degree of the input group, we need to find an n such that n (n-1)/2 > k. This implies for this n that the smallest primitive faithful action of An cannot be realized inside Sk, thus this n is an upper bound for the degree of the alternating group we try to recognise.

To get the formula, bring k on left side (this is the fixed perm degree of input group), and apply pq-formula for polynomial in n. Then the values of n that satisfy the condition on k are given exactly by the inequality n <= 1/2 + sqrt(1/4 + 2k).

gap/generic/SnAnUnknownDegree.gi
FriedrichRober: Only the upper bound N (on the degree of the alternating/symmetric group) requires additional assumptions in the code, the algorithm is a black box algorithm and we exploit in no other place any special properties. However, a bound N is required for the implementation as this determines a (sensible) number of random elements we consider before giving up

gap/generic/SnAnUnknownDegree.gi
fingolfin: Thanks @FriedrichRober that makes sense. I've now updated this comment accordingly.

gap/generic/SnAnUnknownDegree.gi
fingolfin: I've inserted a primitivity check now which will refuse working on imprimitive perm groups.

SoongNoonien (approved):
I'd agree with @TWiedemann that this can finally be merged now.

fingolfin (approved):
Yeah!

PR #273, opened 2021-06-12
Remove convention sections

ssiccha:
- Remove "Conventions" sections from the docs
  - We don't have any conventions. That we only handle finite fields should
    rather go into the introduction.
- Fix a typo
- Document that we can only handle finite fields


fingolfin:
I would actually like to change that we only handle finite fields, and e.g. also handle Z/nZ...

fingolfin (approved):
Still, fine by me, we can change this again later *shrug*

PR #277, opened 2021-06-18
Improve and unify names (2)

ssiccha:
This continues #253.

Renames as follows:
`RIParent` -> `ParentRecogNode`
`methodsforfactor` -> `methodsforimage`
`forfactor` -> `InitDataForImageRecogNode`
`forkernel` -> `InitDataForKernelRecogNode`

The documentation builds on my machine and I've checked a few cross-references and index entries. Everything looks fine there. `testquick.g` also passes on my machine. 

**Edit**: I've moved the whole nice generators business into a separate branch.

### DIFF
--- a/doc/recognition.xml
+++ b/doc/recognition.xml
@@ -218,12 +218,12 @@ The following attributes are defined for recognition nodes:
 <#Include Label="pregensfac">
 <#Include Label="ImageRecogNode">
 <#Include Label="KernelRecogNode">
-<#Include Label="RIParent">
+<#Include Label="ParentRecogNode">
 <#Include Label="fhmethsel">
 <#Include Label="slpforelement">
 <#Include Label="SLPforElement">
 <#Include Label="StdPresentation">
-<#Include Label="methodsforfactor">
+<#Include Label="methodsforimage">
 
 <ManSection>
 <Attr Name="validatehomominput" Arg="ri, x"/>
@@ -402,7 +402,7 @@ recognised by the current recognition node to the image or
 kernel of the found homomorphism:
 
 <ManSection>
-<Attr Name="forkernel" Arg="ri"/>
+<Attr Name="InitialDataForKernelRecogNode" Arg="ri"/>
 <Description>
     This attribute is initialised to a record with only the component
     <C>hints</C> bound to an empty list at the beginning of the
@@ -424,7 +424,7 @@ kernel of the found homomorphism:
 </ManSection>
 
 <ManSection>
-<Attr Name="forfactor" Arg="ri"/>
+<Attr Name="InitialDataForImageRecogNode" Arg="ri"/>
 <Description>
     This attribute is initialised to a record with only the component
     <C>hints</C> bound to an empty list at the beginning of the
@@ -627,7 +627,7 @@ A successful find homomorphism method has the following duties:
     group in the attribute <Ref Attr="Homom"/>. Note that if your homomorphism
     changes the representation (for example going from matrix groups to
     permutation groups), you will have to set the attribute
-    <Ref Attr="methodsforfactor"/> accordingly. Also,
+    <Ref Attr="methodsforimage"/> accordingly. Also,
     <Ref Attr="ValidateHomomInput"/> may be set to a function which returns
     <K>false</K> for values which may cause <Ref Attr="Homom"/> to
     produce the wrong answer, or error.
@@ -652,7 +652,7 @@ A successful find homomorphism method has the following duties:
 <Mark>additional information</Mark>
 <Item>
     A find homomorphism method may store any data into the attributes
-    <Ref Attr="forkernel"/> and <Ref Attr="forfactor"/>, which both are
+    <Ref Attr="InitialDataForKernelRecogNode"/> and <Ref Attr="InitialDataForImageRecogNode"/>, which both are
     records. Components in these
     record that are bound during the recognition will be copied into
     the recognition node of the kernel and image respectively
--- a/doc/renaming_index.xml
+++ b/doc/renaming_index.xml
@@ -1,4 +1,0 @@
-<Index Key="RecognitionInfoFamily"><C>RecognitionInfoFamily</C></Index>
-<Index Key="IsRecognitionInfo"><C>IsRecognitionInfo</C></Index>
-<Index Key="RIFac"><C>RIFac</C></Index>
-<Index Key="RIKer"><C>RIKer</C></Index>
--- a/gap/base/recognition.gd
+++ b/gap/base/recognition.gd
@@ -155,9 +155,9 @@ DeclareAttribute( "ImageRecogNode", IsRecogNode, "mutable" );
 ## <#/GAPDoc>
 DeclareAttribute( "KernelRecogNode", IsRecogNode, "mutable" );
 
-## <#GAPDoc Label="RIParent">
+## <#GAPDoc Label="ParentRecogNode">
 ## <ManSection>
-## <Attr Name="RIParent" Arg="ri"/>
+## <Attr Name="ParentRecogNode" Arg="ri"/>
 ## <Description>
 ##     The value of this attribute is the recognition node of the
 ##     parent of this node in the recognition tree. The top node does not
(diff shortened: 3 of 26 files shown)


FriedrichRober:
I am also fine with calling the inner nodes of a recog tree node a "splitting node".

FriedrichRober:
@ssiccha, there is a file `howtowritearecogmethod.autodoc` where the old names are still used, for example `forfactor`. Is this intended?

FriedrichRober:
@ssiccha, there is a file `howtowritearecogmethod.autodoc` where the old names are still used, for example `forfactor`. Is this intended?

FriedrichRober:
Perhaps `ParentRecogNode` instead of `ParentNodeOfRecogNode` would be more consistent with the names `KernelRecogNode` and `ImageRecogNode`.

ssiccha:
> @ssiccha, there is a file `howtowritearecogmethod.autodoc` where the old names are still used, for example `forfactor`. Is this intended?
> 
> Edit: This is probably not intended, as this page is used in the documentation. Probably we do not grep this file in our renaming script.

I've updated our script. See here:
https://github.com/ssiccha/recog/tree/Improve-Function-Names

ssiccha:
> Perhaps `ParentRecogNode` instead of `ParentNodeOfRecogNode` would be more consistent with the names `KernelRecogNode` and `ImageRecogNode`.

Done.

ssiccha:
> However, the package `AtlasRep` uses the attribute `NiceGens`. So I'll have to add an alias for that and open a PR on `AtlasRep` using the new name.

I've added a synonym to the attribute and the following lines to the documentation of `NiceGenerators`, see the fixup commit:
> `<P/>`
> For backwards compatibility, the name <Ref Attr="NiceGens"/> is a
> synonym for <Ref Attr="NiceGenerators"/>. This will be deprecated in the
> near future.

ssiccha:
Btw.: I'm seeing whether I can come up with better names for the whole `CalcNice...` business.

contrib/akos/examples4.g
```diff
@@ -485,7 +485,7 @@ if d>18 and d mod 16 = 2 and q mod 4 = 3 then
       od;
       list := reverse(Set(list));
       return list;
-  fi;  # largest prime factor of d-2
+  fi;  # largest prime image of d-2
```
wilfwilson, line 488: The dangers of search and replace 😉 
```suggestion
  fi;  # largest prime factor of d-2
```

doc/recognition.xml
```diff
@@ -466,7 +466,7 @@ This attribute returns a function that compares two elements of the
 group being recognised.  Usually this is
 just the operation <Ref Oper="EQ" BookName="Ref"/> but for projective
 groups it is a special function checking for equality up to a scalar
-factor. In
+image. In
```
wilfwilson, line 469: Also should be factor.
```suggestion
factor. In
```

misc/obsolete/semilinear.gi
```diff
@@ -136,7 +136,7 @@ FindHomMethodsProjective.NotAbsolutelyIrred := function(ri,G)
   fi;
 
   Info(InfoRecog,2,"Rewriting generators over larger field with smaller",
-       " degree, factor=",MTX.DegreeSplittingField(m));
+       " degree, image=",MTX.DegreeSplittingField(m));
```
wilfwilson, line 139: I'm _pretty_ sure this should also remain as factor.
```suggestion
       " degree, factor=",MTX.DegreeSplittingField(m));
```

misc/steve/obrienmalle.gi
```diff
@@ -1,7 +1,7 @@
 # GAP translation attempt Jan 2004 SL
 
 #**************************************************************/
-#   Identify the non-abelian simple composition factor        */
+#   Identify the non-abelian simple composition image        */
```
wilfwilson, line 4:
```suggestion
#   Identify the non-abelian simple composition factor        */
```

tst/working/quick/ProjSubfield2.tst
```diff
@@ -5,6 +5,6 @@ gap> x := PseudoRandom(GL(6,5^3));;
 gap> gens := List(gens,y->y^x);;
 gap> g := GroupWithGenerators(gens);;
 
-# Remark: The *6 comes from the fact that by blowing up we lose a factor
+# Remark: The *6 comes from the fact that by blowing up we lose a image
```
wilfwilson, line 8:
```suggestion
# Remark: The *6 comes from the fact that by blowing up we lose a factor
```

wilfwilson (changes requested):
Sergio! I'm happy with these four name changes, but you have been a bit overzealous with the search-and-replace of "factor" to "image".

contrib/akos/examples4.g
ssiccha: Haha, I thought I checked them all but I must have missed some! :D Thanks for finding them!

tst/working/quick/ProjSubfield2.tst
ssiccha: Huh, strange. I'm pretty sure I reverted that replacement. Maybe I messed up some git stuff? 🤔 Anyways, thanks again. :)

fingolfin:
This needs to be rebased first?

wilfwilson (approved):
Thanks for making those fixes. But yes, it's rebasing time 🙂

ssiccha:
Rebased onto master. Let's see whether the test workflows pass.

doc/renaming_index.xml
```diff
@@ -2,3 +2,7 @@
 <Index Key="IsRecognitionInfo"><C>IsRecognitionInfo</C></Index>
 <Index Key="RIFac"><C>RIFac</C></Index>
 <Index Key="RIKer"><C>RIKer</C></Index>
+<Index Key="RIParent"><C>RIParent</C></Index>
+<Index Key="methodsforfactor"><C>methodsforfactor</C></Index>
+<Index Key="forfactor"><C>forfactor</C></Index>
+<Index Key="forkernel"><C>forkernel</C></Index>
```
fingolfin, line 8: Not about this PR, but: why is there a separate index file? Why aren't those index entries directly in `renaming_table.xml`?

gap/base/recognition.gd
```diff
@@ -155,17 +155,17 @@ DeclareAttribute( "ImageRecogNode", IsRecogNode, "mutable" );
 ## <#/GAPDoc>
 DeclareAttribute( "KernelRecogNode", IsRecogNode, "mutable" );
 
-## <#GAPDoc Label="RIParent">
+## <#GAPDoc Label="ParentRecogNode">
```
fingolfin, line 158: These renames potentially break existing code using recog.

Is there a strong reason why we don't use `DeclareObsoleteSynonym` resp. `DeclareObsoleteSynonymAttr` for these (and other, previously renamed things)?

doc/recognition.xml
```diff
@@ -402,7 +402,7 @@ recognised by the current recognition node to the image or
 kernel of the found homomorphism:
 
 <ManSection>
-<Attr Name="forkernel" Arg="ri"/>
+<Attr Name="InitDataForKernelRecogNode" Arg="ri"/>
```
fingolfin, line 405: I was confused by this name as to me `init` stands for the verb `initialize`, not for `initial` or `initialization`. And that makes for some confusing reading. So perhaps
```suggestion
<Attr Name="InitialDataForKernelRecogNode" Arg="ri"/>
```

doc/renaming_index.xml
ssiccha: I changed our script such that the index is also written into the `renaming_table.xml`.

ssiccha: And I also tested whether the index still works. Everything is fine.

gap/base/recognition.gd
ssiccha: I added synonyms for the renames in this PR and am currently opening a PR for the renames that are already in master.

doc/recognition.xml
ssiccha: I changed the names accordingly.

ssiccha:
I've double-checked the touched files and the links  in the generated manual. Everything looks fine so I'm merging this.

PR #282, opened 2021-08-11
Refactor FastNormalClosure and don't use it when kernel is abelian

ssiccha:
- Refactor FastNormalClosure
- Don't use FastNormalClosure if kernel is abelian


gap/generic/kernel.gi
```diff
@@ -77,36 +77,35 @@ end );
 # Computes randomly (it might underestimate) the normal closure of <list>
 # under conjugation by the group generated by <grpgens>.
```
danielrademacher, line 78: Maybe add some information about <n> from the input?

ssiccha: Jepp, good point.

gap/matrix.gi
```diff
@@ -539,8 +540,9 @@ function(ri,G)
   Setmethodsforfactor(ri,FindHomDbProjective);
 
   # the kernel:
+  # The kernel is abelian, so we don't need to do normal closures.
+  findgensNmeth(ri).func := FindKernelRandom;
```
danielrademacher, line 544: As above.

gap/matrix.gi
```diff
@@ -239,8 +239,9 @@ function(ri, G)
       else   # exactly two blocks:
           # FIXME: why don't we just compute a precise set of generators of the kernel?
           # That should be easily and efficiently possible at this point, no?
+          # The kernel is abelian, so we don't need to do normal closures.
+          findgensNmeth(ri).func := FindKernelRandom;
```
danielrademacher, line 243: Is "findgensNmeth(ri).func" correct? I have not found it at any other point in the package. Maybe "findgensNmeth(ri).method"?

gap/matrix.gi
```diff
@@ -239,8 +239,9 @@ function(ri, G)
...
           findgensNmeth(ri).args[1] := 7;
-          findgensNmeth(ri).args[2] := 5;
```
danielrademacher, line 243: I'm also not sure, what arguments of which function are stored here. Maybe some other reviewer has already worked with this?

gap/matrix.gi
```diff
@@ -539,8 +540,9 @@ function(ri,G)
   Setmethodsforfactor(ri,FindHomDbProjective);
 
   # the kernel:
+  # The kernel is abelian, so we don't need to do normal closures.
+  findgensNmeth(ri).func := FindKernelRandom;
   findgensNmeth(ri).args[1] := Length(ri!.blocks)+10;
-  findgensNmeth(ri).args[2] := 7;
```
danielrademacher, line 543: As above.

danielrademacher (commented):
The function "FastNormalClosure" looks very good. I like the changes.

I also checked that the kernel is abelian and, hence, we don't need to call the normal closure function in the lines improved by @ssiccha.

gap/generic/kernel.gi
FriedrichRober: I second that. It would be nice to know what the error probability is in terms of `n` (and the maximal length of subgroup chains in `<grpgens>`).

gap/matrix.gi
FriedrichRober: Well spotted. The documentation says that `findgensNmeth(ri)`  is a record with at least two components bound, namely `method` and `args`.

ssiccha: Nice. That code must have been called by the tests I think. So the docs may be outdated. I'll double check when I'm back from holidays next week.

gap/matrix.gi
ssiccha: That should be explained in the docs of findgensNmeth. I'll check.

gap/matrix.gi
ssiccha: Oh, or it simply never got called since I put it into the wrong component. ^^

fingolfin (changes requested):
This looks plausible but there are things that were pointed out in review that still need to be addressed.

gap/matrix.gi
ssiccha: These are the arguments to the function stored in `findgensNmeth(ri).method`, in this case `FindKernelRandom`.

ssiccha:
I added a comment referencing Lemma 2.3.8 from "Permutation Group Algorithms" by Akos Seress, which reads:
> **Lemma 2.3.8.** Suppose that `G = <S>` acts on H and let U ⊆ H . Let g be a
> random subproduct of S and u be a random subproduct of U . If <U > is not
> closed for the G-action then Prob(u ^ g not in <U >) ≥ 1/4.

I addressed all comments and added two small commits:
- Add small optimisation to FastNormalClosure
- Fix documentation of findgensNmeth

ssiccha:
Rebased onto master.

danielrademacher (approved):
I checked everything again and this looks great now. Thanks @ssiccha.

fingolfin (approved):

gap/generic/kernel.gi
```diff
@@ -75,38 +75,46 @@ InstallGlobalFunction( RandomSubproduct, function(a)
...
-        randlist:=RandomSubproduct(list2);
-      fi;
-      if not IsOne(randlist) then
-         for randgens in grpgens do
-             Add(list2, randlist^randgens);
-         od;
+  local list2, fewGenerators, repetitions, randlist, conjugators, i, c;
+  list2 := ShallowCopy(list);
+  if Length(grpgens) = 1 then
+    return list2;
```
fingolfin, line 89: Note that if we instead of `gens` we passed a group object, we could test here for `HasIsAbelian(G) and IsAbelian(G)`. And in all places calling here that I found, there already was such a group object -- with the exception of a call in `IsThisSL2Natural`. But there in turn, all (?) callers to `IsThisSL2Natural` could easily be changed to pass on a group object.

One could even support both variants in this function, by turning `gens` into a local var, and calling the first argument `G`, and doing something like `if IsGroup(G) then gens := GeneratorsOfGroup(G); else gens:=G; fi;` or so

ssiccha: I'll put this idea into its own issue.

PR #284, opened 2021-08-19
Add argument check to FastNormalClosure

ssiccha:
Builds on #282 and contains one more commit, namely:
- Add argument check to FastNormalClosure

Add a check to make sure whether the argument controlling how often conjugates are
computed is a positive integer. There were a few situations where it was set to 0.


ssiccha:
@fingolfin Tests pass on my machine now.

ssiccha:
Conflicts again

ssiccha:
Rebased.

gap/projective/c3c5.gi
```diff
@@ -178,7 +178,8 @@ function(ri, G)
   # There might be a kernel, because we have more scalars over the bigger
   # field, so go for it, however, fewer generators should suffice:
   # Also, doing normal closure will not help!
-  findgensNmeth(ri).args := [5,0];
+  findgensNmeth(ri).method := FindKernelRandom;
+  findgensNmeth(ri).args := [5];
```
danielrademacher, line 182: I only know in theory what this is supposed to do, but I don't know enough about the package to understand what this change does. Is "FindKernelRandom" the default method and you're clarifying that here?

ssiccha: It's not the default method. The default method additionally does normal closure computations, that is it also takes conjugates of the found kernel generators. Here we don't do conjugations, though. Previously this was done by passing a 0 as second argument to the default method. Now, I simply set the method which does not do kernel computations.

danielrademacher: Ah, thanks for explaining!

PR #285, opened 2021-08-20
Move tests from .../quick/ to .../slow/

ssiccha:
Some tests got a bit out of hand and now take upwards of 60s to finish. I want
to frequently and quickly check whether I broke something obvious, and then I'd
rather have the quick suite finish in under 60s in total.


ssiccha:
I didn't remove any tests from the test suite as a whole. The tests in `.../slow` are also run by our GitHub Actions setup. Please double-check whether I accidentally removed a whole file or something like that.

danielrademacher (approved):
Looks good and I like the idea. 👍

Maybe one can add a function to execute all fast tests, all fast and slow tests and all fast, slow and very slow tests?

tst/working/quick/ClassicalNaturalNaming.tst
```diff
@@ -9,59 +9,30 @@ true
 #
 # linear groups
 #
-gap> d:=3;; for q in [2, 3, 4, 5, 7, 8, 9, 11, 13] do TestNaming("SL", d, q); od;
-gap> d:=4;; for q in [2, 3, 4, 5, 7, 8, 9, 11, 13] do TestNaming("SL", d, q); od;
-gap> d:=5;; for q in [2, 3, 4, 5, 7, 8, 9, 11, 13] do TestNaming("SL", d, q); od;
+gap> d:=3;; for q in [2, 3] do TestNaming("SL", d, q); od;
```
fingolfin, line 12: This makes me unhappy (which is not saying we shouldn't merge it), because (a) the naming should be FAST and if we have to drop this test, it  kinda means that it is not; and (b) now two very similar tests exist in multiple files, and when one looks at only one of them, it may be easy to miss that it has a "sibling" elsewhere.

So for now, we may have to use this patch, but I'd prefer to address (a) by making naming faster and (b) e.g. by adding comments to the "sibling" files mentioning the existence of the respective other. An alternative (and perhaps nicer) fix for (b) would be to introduce a global variable that indicates whether we are running quick/normal/slow/veryslow tests; and then *the same file* performs a different subset of tests depending on this setting. That way, the "alternate" tests are right next to each other. There are different ways to achieve this; one would be to use the `#@if` syntax Chris introduced some time ago.

ssiccha: Yeah, I think it makes sense to clean that up a bit. I'll address this in a new PR.

ssiccha: See #287. Understanding how to make the test pragmas work took me quite a while since I found them very confusing.. I submitted an issue to gap-system/gap for that.

PR #293, opened 2021-09-17
Classical groups. Express elements in LGO (nice) generators.

danielrademacher:
Include functions to express arbitrary elements in terms of LGO generators. We use the LGO generators as "nice" generators for the constructive recognition of classical groups.


fingolfin (changes requested):
I am not keen on adding binaries like PDFs into this repository. There is also the copyright issue (I have not yet looked at them). Also `.DS_Store` files should not be in there (perhaps add those to a global `.gitignore` list, on Mac's they can otherwise creep in quite easily.

PR #294, opened 2021-10-05
Improve veryslow/MatFDPM.tst

ssiccha:


AnnaKDS (approved):

tst/working/veryslow/MatFDPM.tst
```diff
@@ -1,33 +1,42 @@
...
+> pos := Position(dims,max);
+> x := PseudoRandom(GL(max,field));
+> mgens := List(cf[pos].generators,y->y^x);
+> h := Group(mgens);
+> return Size(RecogniseGroup(h)) = Size(g);
+> #return RECOG.TestGroup(h,false,Size(g));
+> end;;
+gap> l := [];
+gap> for deg in [10, 13, 17, 32] do
+> for field in List([2,3,4,7,9], GF) do
```
FriedrichRober, line 25: Why have you removed the fields of size 5 and 11?

tst/working/veryslow/MatFDPM.tst
```diff
@@ -1,33 +1,42 @@
...
+> Add(l, testFDPM(deg, field, false));
+> Add(l, testFDPM(deg, field, true));
+> od;
+> od;
+> return;
+> end;;
+
+gap> testFDPM(88, GF(3), false);
+gap> testFDPM(88, GF(2), true);
```
FriedrichRober, line 42: Same as above. These lines throw an error, since the output should be
```
F dim=  87 field=3  0     
true

F dim=  86 field=2  0     
true
```

tst/working/veryslow/MatFDPM.tst
```diff
@@ -1,33 +1,42 @@
...
+> Add(l, testFDPM(deg, field, true));
+> od;
+> od;
+gap> func := function(l)
+> for deg in [10, 13, 17] do
+> for field in List([2,3,4,7,9], GF) do
+> Add(l, testFDPM(deg, field, false));
+> Add(l, testFDPM(deg, field, true));
+> od;
+> od;
```
FriedrichRober, line 37: The loops throw an error on my local machine, since the expected output are lines of the format:
```
 F dim=   8 field=2  0      0 
```

tst/working/veryslow/MatFDPM.tst
```diff
@@ -1,33 +1,42 @@
...
+> dims := List(cf,x->x.dimension);
+> max := Maximum(dims);
+> pos := Position(dims,max);
+> x := PseudoRandom(GL(max,field));
+> mgens := List(cf[pos].generators,y->y^x);
+> h := Group(mgens);
+> return Size(RecogniseGroup(h)) = Size(g);
+> #return RECOG.TestGroup(h,false,Size(g));
+> end;;
+gap> l := [];
```
FriedrichRober, line 23:
```suggestion
gap> l := [];;
```

FriedrichRober (changes requested):

tst/working/veryslow/MatFDPM.tst
ssiccha: Because I think the old tests were a bit excessive. I just removed a bit here and there. We'll have to double check with the codecov bot whether this leads to us covering less code.

ssiccha: I'll have to resolve the conflicts first.

fingolfin (commented):
Maybe clarify in the commit message what this improves: namely before, executing this tst file would just test some random subset of all combinations things it could test; while now it systematically tests multiple of these combinations. (assuming I am guessing this right?)

Otherwise it seems plausible to me

tst/working/veryslow/MatFDPM.tst
fingolfin: Need to reduce the info level temporarily to get rid of these (they are randomized so that's really the only option)

ssiccha:
I improved the commit message and cleaned up the file. Now it actually does something and I get "incorrect size" errors.

ssiccha:
This should be done now @FriedrichRober. Let's see whether the tests pass. Once the tests passed, let's double check that code coverage did not decrease.

Also now avoid the spurious error that the random generating set we
choose only generated a proper subgroup of the one we wanted to test.

FriedrichRober (approved):
Looks good to me. I checked code coverage and we only gained new hits, so this is nice.

PR #303, opened 2021-12-17
Improve printing of recog trees

ssiccha:
Slightly change how nodes are printed. Previously, the name of the
successful method might be appended by a `comment` string. Now, if
there is a comment, it is printed in a separate `Comment:` field.

Thus also remove the leading underscore of the `ri!.comment` values.

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -43,7 +43,7 @@ RECOG_ViewObj := function( level, ri )
             Print(ms);
         fi;
         if IsBound(ri!.comment) then
-            Print(ri!.comment);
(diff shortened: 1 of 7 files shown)


ssiccha:
I'll turn Comment : into `Comment=` for uniformity. Also I need to adjust the test files a little. My son just woke up though. If anyone wants to, feel free to finish this in the mean time.

ssiccha:
I adjusted the commit. Tests pass on my machine. Let's see.

danielrademacher (approved):
Looks very good. Thanks for the improvement und the idea!

Issue #288, opened 2021-08-23
Let FastNormalClosure take a group as first argument


ssiccha:
Note that if we instead of `gens` we passed a group object, we could test here for `HasIsAbelian(G) and IsAbelian(G)`. And in all places calling here that I found, there already was such a group object -- with the exception of a call in `IsThisSL2Natural`. But there in turn, all (?) callers to `IsThisSL2Natural` could easily be changed to pass on a group object.

One could even support both variants in this function, by turning `gens` into a local var, and calling the first argument `G`, and doing something like `if IsGroup(G) then gens := GeneratorsOfGroup(G); else gens:=G; fi;` or so

_Originally posted by @fingolfin in https://github.com/gap-packages/recog/pull/282#r693956976_

ssiccha:
Fixed by #304.

PR #304, opened 2021-12-21
Improve FastNormalClosure

ssiccha:
Add the RecogNode as argument.

Drop the undocumented assumption, that the group generated by <grpgens>
is an overgroup of the one generated by <list>. Thus change a test
whether the former group is cyclic into a test whether the entries of
<grpgens> commute with the ones in <list>.


ssiccha:
I could combine this with the issue in #288. But in some situations the generators of the group which are taken do not come with a RecogNode. I'll have to think about this again.

ssiccha:
I got rid of the whole recog node shenanigans. This is ready for review now.

FriedrichRober (approved):
Ok, looks good to me.

PR #305, opened 2021-12-29
misc: add a comment

ssiccha:


gap/base/recognition.gi
```diff
@@ -854,6 +854,9 @@ InstallGlobalFunction( "SLPforNiceGens", function(ri)
   return s;
 end );
 
+# For user debugging purposes. Takes a string consisting of lower- or
+# upper-case "f" and "k". Returns the node one arrives at by descending through
+# the tree, going to the factor for each "f" and to the kernel for each "k".
```
danielrademacher, line 859: What happens if one tries a non existing path?

danielrademacher (approved):

gap/base/recognition.gi
ssiccha: Either ImageRecogNode and KernelRecogNode should error.  Since it's a debugging function that's good enough IMO. If they are worth it, We could move all of those debug functions into a separate file, and provide short documentations.

ssiccha: NVM, that may be the only function that is only used for debugging.

danielrademacher: Ok, thanks! 👍

PR #306, opened 2022-01-05
Clarify docs of IsReady

ssiccha:
The documentation was very vague. Write into the documentation what the
package currently does with the flag IsReady.


FriedrichRober (approved):
This is much better explained than before. Thank you.

PR #308, opened 2022-01-10
Improve RecogniseGeneric and document RecogNode

ssiccha:
- Improve documentation of IsReady
- Improve RecogniseGeneric
  - Clean up of comments.
  - Some refactoring.
  - Improve the warning message if recognition of a node does not
    terminate. Since IsReady is not set, an error will be thrown once one
    wants do something with the recognition data structure. Thus it is not
    necessary to throw an error in RecogniseGeneric. Make
    sure that, if everything went fine, then IsReady is set directly
    before returning.
- Rename and document recognition node constructor
  - Also rename the corresponding type.
    Rename as follows:
    RecognitionInfoType -> RecogNodeType
    EmptyRecognitionInfoRecord -> RecogNode

### DIFF
--- a/gap/base/recognition.gi
+++ b/gap/base/recognition.gi
@@ -129,7 +129,7 @@ InstallGlobalFunction( RecogniseGroup,
 InstallGlobalFunction( TryFindHomMethod,
   function( g, method, projective )
     local result,ri;
-    ri := EmptyRecognitionInfoRecord(rec(),g,projective);
+    ri := RecogNode(g,projective);
     Unbind(g!.pseudorandomfunc);
     result := method(ri,g);
     if result in [TemporaryFailure, NeverApplicable] then
@@ -141,11 +141,13 @@ InstallGlobalFunction( TryFindHomMethod,
     fi;
   end );
 
-InstallGlobalFunction( EmptyRecognitionInfoRecord,
-  function(r,H,projective)
+InstallMethod( RecogNode,
+  "standard method",
+  [ IsGroup, IsBool, IsRecord ],
+  function(H,projective,r)
     local ri;
     ri := ShallowCopy(r);
-    Objectify( RecognitionInfoType, ri );
+    Objectify( RecogNodeType, ri );
     SetGrp(ri,H);
     Setslpforelement(ri,SLPforElementGeneric);
     SetgensN(ri,[]);       # this will grow over time
@@ -212,6 +214,20 @@ InstallGlobalFunction( EmptyRecognitionInfoRecord,
     return ri;
   end );
 
+InstallOtherMethod( RecogNode,
+  "for a group",
+  [ IsGroup ],
+  function(H)
+    return RecogNode(H, false, rec());
+  end );
+
+InstallOtherMethod( RecogNode,
+  "for a group and a boolean",
+  [ IsGroup, IsBool ],
+  function(H, projective)
+    return RecogNode(H, projective, rec());
+  end );
+
 # Sets the stamp used by RandomElm, RandomElmOrd, and related functions.
 RECOG.SetPseudoRandomStamp := function(g,st)
   if IsBound(g!.pseudorandomfunc) then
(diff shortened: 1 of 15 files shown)


ssiccha:
@FriedrichRober Since you also looked at the last PR concerning the docs of IsReady. Could you have a look at the updated documentation of IsReady again and tell me whether you think that's again an improvement?

gap/base/recognition.gi
```diff
@@ -555,7 +582,7 @@ InstallGlobalFunction( RecogniseGeneric,
         fi;
 
         if not IsReady(rifac) then
-            # the recognition of the image failed, also give up here:
+            # IsReady is not set, thus abort the whole computation.
             if InfoLevel(InfoRecog) = 1 and depth = 0 then Print("\n"); fi;
```
FriedrichRober, line 586: Should we print here some information that the computation is aborted?

```suggestion
            if InfoLevel(InfoRecog) = 1 and depth = 0 then Print("\n"); fi;
            Info(InfoRecog, 1,
                 "ImageRecogNode of node <ri> could not be recognised,",
                 " IsReady(<ri>) is not set!");
```

gap/base/recognition.gi
```diff
@@ -625,20 +652,22 @@ InstallGlobalFunction( RecogniseGeneric,
         SetParentRecogNode(riker,ri);
         Info(InfoRecog,2,"Back from kernel (depth=",depth,").");
 
+        if not IsReady(riker) then
```
FriedrichRober, line 655: Here the same, maybe we should print something here?

FriedrichRober (approved):
I only have some minor comments. The reworked documentation for `IsReady` is fine.

gap/base/recognition.gi
ssiccha: Good point. I'll add such an info message to the "unfinished case". And now that I think about it, we don't need to print anything in the "image" or "kernel" parts of RecogniseGeneric, since the warning message is printed in the "unfinished case" part already.

gap/base/recognition.gi
ssiccha: We don't need to print something here, since we're only "exiting". The warning message was already printed in one of the "failed cases" that is in the "unfinished case" part or in the "recognise the image" part.

ssiccha:
@FriedrichRober I think I have again made the docs of IsReady a bit more precise. Could you have another quick look? ^^

ssiccha:
Ok, now the docs of IsReady should be good to go. The version before the `Fix docs of IsReady` should be correct once Mandarins are implemented. But we'll see that once the Mandarins are actually finished.

FriedrichRober:
The doc of IsReady sounds now a bit more technical, but I understand what is going on. I approve.

gap/base/recognition.gi
```diff
@@ -212,6 +214,27 @@ InstallGlobalFunction( EmptyRecognitionInfoRecord,
...
+
+InstallOtherMethod( RecogNode,
+  "for a group and a boolean",
+  [ IsGroup, IsBool ],
+  function(H, projective)
+    return RecogNode(H, projective, rec());
+  end );
+
+InstallOtherMethod( RecogNode,
+  "for backwards compatibility",
```
fingolfin, line 232: What are you trying to be backwards compatible with, though? is there still code in the repo that calls it like that? If so, can't we just update that code?

Other than that: any external code won't be calling `RecogNode`, as it didn't exist before, right? It'll call `EmptyRecognitionInfoRecord`, so that code needs to be adjusted anyway. Or perhaps we can retain `EmptyRecognitionInfoRecord` in a new file `gap/obsolete.gi`; but then that backwards compatible function can take of reordering the arguments.

Perhaps I am missing something, though?

fingolfin (approved):
Looks good to me (but note my comment on the backwards compatibility method)

gap/base/recognition.gi
ssiccha: > Perhaps I am missing something, though?

Nope, that makes perfect sense, thanks! :+1:

ssiccha: I'm going with this:
> Or perhaps we can retain EmptyRecognitionInfoRecord in a new file gap/obsolete.gi; but then that backwards compatible function can take of reordering the arguments.

ssiccha:
I had to fiddle around a bit with the backwards compatibility method, but now it works on my machine. I'll wait for tests to pass, then this gets merged.

PR #309, opened 2022-01-27
CI: Ensure code coverage reporting will work in February 2022 and beyond

fingolfin:
See https://github.com/gap-actions/process-coverage/issues/10


ssiccha:
Building the manuals errors with:
`Error, a different version of recog was already loaded or you started GAP in the wrong directory; try `gap -A regen_doc.g` or `gap -A makedoc.g` called from`

wilfwilson (approved):

PR #310, opened 2022-02-01
Fix docs of CalcNiceGensHomNode

ssiccha:


ssiccha:
[This](https://github.com/gap-packages/recog/blob/d67cd5d214bdfe5a57472125bc58f1f2646be9ca/gap/base/recognition.gi#L660) is what actually happens, which is not what is decribed in the master-branch version of the docs.

danielrademacher (approved):

PR #346, opened 2025-12-08
Improve README to make installation easier

SoongNoonien:
Yesterday, while installing this package I had a few problems getting it to work. I'm quite sure that this was at least partly my fault but I think this change might help to reduce future problems for other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ To enable GAP to find recog you can start GAP via
 
 > [!TIP]
 > You can add this option to your GAP startup script. More information about
-> this can be found [in Chapter 3 of the GAP reference manual](https://docs.gap-system.org/doc/ref/chap3.html).
+> this can be found in [Chapter 3 of the GAP reference manual](https://docs.gap-system.org/doc/ref/chap3.html).
 
 ## Loading
 


README.md
```diff
@@ -8,31 +8,52 @@ A GAP package for group recognition
...
+If you want to contribute to recog you might want to install it
+into a custom directory. For example into the directory `~/packages`. Then you
+can clone recog (or perhaps your personal fork of it) into `~/packages` via
+
+    cd ~/packages
+    git clone git@github.com:gap-packages/recog.git
+
+To enable GAP to find recog you can start GAP via
+
+    gap --packagedirs "~/packages;"
```
fingolfin: Note that this is a very new option, IIIRC added in GAP 4.14 (but we should check GAP's `CHANGES.md`).

So I think this should be listed as a warning.

let's talk about this later

SoongNoonien: I've checked and it was introduced in GAP 4.15 in September. This is the respective [PR](https://github.com/gap-system/gap/pull/5873).

README.md
```diff
@@ -8,31 +8,60 @@ A GAP package for group recognition
...
+
+> [!IMPORTANT]
+> Note that `--packagedirs` was introduced in GAP 4.15. For older version please
+> refer to [Chapter 76 of the GAP reference manual](https://docs.gap-system.org/doc/ref/chap76.html)
+> for alternative solutions.
+
+> [!TIP]
+> You can add this option to your GAP startup script. More information about
+> this can be found [in Chapter 3 of the GAP reference manual](https://docs.gap-system.org/doc/ref/chap3.html).
```
fingolfin:
```suggestion
> this can be found in [Chapter 3 of the GAP reference manual](https://docs.gap-system.org/doc/ref/chap3.html).
```

PR #366, opened 2025-12-12
Remove `contrib/alice/classical.g`

fingolfin:
This seems to be a predecessor of the naming code in `gap/matrix/classical.gi` by @aniemeyer -- but it is rather different. Perhaps there are bits and pieces worth saving? Otherwise, let's just remove it.


aniemeyer:
This is very old code and can be delted.

PR #330, opened 2024-06-19
Include constructive recognition sl

danielrademacher:

Issue #237, opened 2021-02-15
Add RandomOrd and extend RandomElm[Ord]


ssiccha:
In the discussion for #197 @fingolfin wrote:

> In the two existing use cases for RandomElmOrd we actually only use the order part. So perhaps we should change it to only return the order, not a record with other stuff we just discard.

> Overall RandomElmOrd has quite some overhead. So an additional idea would be to add an argument N which indicates how many random elements we want, and then return a list of the corresponding orders.

PR #371, opened 2026-02-03
Add constructive recognition for SL_n, take 2

fingolfin:
This is a successor for PR #330. The only difference for now is that the branch is now hosted in this repo, not in @danielrademacher's branch. This will make it easier for us here to work on it together.

The plan is to further reduce diffs between this and master, and then get it merged.

There will also soon be another PR / branch with the rest of Daniel's code.

TODOs for merging this:
- no merge conflicts
- all tests must pass
- understand the change in `gap/projective/classicalnatural.gi‎` and possibly remove or adjust it (see also PR #487)
- ensure the code added here has good code coverage (i.e., is actually used)
  - we can keep some "dead" code if there is a good reason, but then it should be commented accordingly, otherwise it'll lead to a lot of confusion
  - if there are no immediate plans to use the dead code, we should consider instead putting it into a `contrib` subdir (as part of this PR)
- generally perform a code review and make sure we understand everything
- look out for bottlenecks and potential optimizations (but: correctness over performance, always)


gap/projective/classicalnatural.gi
```diff
@@ -654,7 +654,8 @@ RECOG.IsThisSL2Natural := function(gens,f)
       Info(InfoRecog,4,"SL2: Computing stabiliser chain.");
       S := StabilizerChain(Group(gens));
       Info(InfoRecog,4,"SL2: size is ",Size(S));
-      return Size(S) mod (q*(q-1)*(q+1)) = 0;
+      # return Size(S) mod (q*(q-1)*(q+1)) = 0;
+      return Size(S) = (q*(q-1)*(q+1));
```
fingolfin, line 654: This change is fishy. See also https://github.com/gap-packages/recog/pull/487

I wonder why we need it. But if we do need it, then certainly there must be more; e.g. the code comment a few lines above references this code and thus must be adjusted.

fingolfin:
```suggestion
      return Size(S) mod (q*(q-1)*(q+1)) = 0;
```

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+# TODO: Add MSLPs to these algorithms
+# This algorithm is an implementation of the paper
+# "Fast Recognition of Classical Groups over Large Fields"
+# by Marston Conder and Charles Leedham-Green
+RECOG.ConstructiveRecognitionSL2NaturalRepresentation := function(G, q, epsilon)  # TODO: never called
+local F, factors, counter, one, factor, foundEle, passed, max, eigenvalues, eigenvalues2, eigenvectors, eigenvectors2, A, B, rand, foundConjugate, test, BDiag, C, basechange, v, d, CC, check, i, j, D, DD, T, S, o, t1, t2, zero, slp;
+
+    factors := PrimeDivisors(q-1);
+    counter := 1;
+    F := GF(q);
```
fingolfin, line 35: For this (and all other future code) we should strive to pass around the field, not the size.

This is key first step towards supporting alternative field implementations (such as those from the `StandardFF` package).

Though in this case, really the field should be derived from `G` (i.e. the group `G` is generated by matrices which have a certain `BaseDomain`...). On the GAP side we are [discussing adding a `BaseDomain` for matrix groups](https://github.com/gap-system/gap/issues/6389).

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+
+    if not(IsObjWithMemory(GeneratorsOfGroup(G)[1])) then
+        G := GroupWithMemory(G);
+    fi;
+
+    one := One(G);
```
fingolfin: Better to use `IsOne` (it used to be that `IsOne` was super slow for matrices, but this was fixed some years ago)
```suggestion
```

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+        #            counter := counter + 1;
+        #        fi;
+        #    else
+        #        counter := counter +1;
+        #    fi;
+        #    if counter >= max then
+        #        return fail;
+        #    fi;
+        #od;
+
```
fingolfin, line 69: Any reason to keep this? Otherwise I'd rather delete this.
```suggestion
```

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+        # Step 1: Find random element of order q-1 (first improvement based on magma code)
```
fingolfin, line 70:
```suggestion
        # Step 1: Find random element of order q-1
```
I don't mind crediting Magma, but this comment is super vague. If anyone knows what improvements this refers to...?

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+            if (o mod (q - 1) = 0) then
+                A := A^(o/(q - 1));
+                foundEle := true;
+            else
+                counter := counter +1;
+            fi;
+            if counter >= max then
+                return fail;
+            fi;
+        od;
```
fingolfin, line 84: can be simplified
```suggestion
        while true do
            A := PseudoRandom(G);
            o := Order (A);
            if (o mod (q - 1) = 0) then
                A := A^(o/(q - 1));
                break;
            fi;
            counter := counter +1;
            if counter >= max then
                return fail;
            fi;
        od;
```

Perhaps we should also use `RandomElmOrd` or so (see also #237, #462)

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+###################################################################################################
+###################################################################################################
+
+
+
+# TODO: Add MSLPs to these algorithms
+# This algorithm is an implementation of the paper
+# "Fast Recognition of Classical Groups over Large Fields"
+# by Marston Conder and Charles Leedham-Green
+RECOG.ConstructiveRecognitionSL2NaturalRepresentation := function(G, q, epsilon)  # TODO: never called
```
fingolfin, line 30: I started writing comments on this code before realizing: This is never called *sigh*

But I decided to leave my comments anyway: while we may not wish to apply them, and instead delete this function, I feel that they show some general patterns of things that ought to be improved.

And of course we *could* give this code a try and compare it to the code by Frank which @Till-Eisen recently integrated? Till, perhaps you'd be interested in looking into that a bit (don't spend too much time, in particular if it turns out difficult to run this code)

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+
+    return fail;
+
+end;
+
+
+
+# Note that we use the discrete logarithm to normalise the primitive element at position [1,1]. But this is not necessarly as the entry at position [1,1] is primitive.
+# Hence, this function can be adapted to larger fields by avoiding the normalisation step
+RECOG.ConstructiveRecognitionSL2NaturalRepresentationCompleteBasis := function(list,F,q,p,f)  # TODO: never called
```
fingolfin, line 210: keep or delete?

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+        g := PseudoRandom(G);
+        order := Order(g);
+        if (order mod 2 = 0) then
+            h := g^(order/2);
+            EPlus := RECOG.FixspaceMat(h);
+            if Size(EPlus) = 2 then
+                foundStrongInvoluation := true;
+            fi;
+        fi;
+    od;
```
fingolfin, line 374:
```suggestion
    local d, G, g, h, order, gensCentraliser, EPlus;
    
    d := 4;
    G := GroupByGenerators(X);
    while true do
        g := PseudoRandom(G);
        order := Order(g);
        if order mod 2 = 0 then
            h := g^(order/2);
            EPlus := RECOG.FixspaceMat(h);
            if Size(EPlus) = 2 then
                break;
            fi;
        fi;
    od;
```

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+    gensCentraliser := RECOG.InvolutionCentraliser(G,h,100);
+    # TODO: Compute generating set for OmegaX(E) (see paper from LGO)
+
+    #TODO: CONTINUE HERE
+
+end;;
+
+
+
+# Input: h an involuation in a BB group G, natural number N > 0
```
fingolfin, line 385:
```suggestion
# Input: h an involution in a BB group G, natural number N > 0
```

gap/projective/constructive_recognition/SL/BaseCase.gi
```diff
@@ -0,0 +1,413 @@
...
+            if Size(EPlus) = 2 then
+                foundStrongInvoluation := true;
+            fi;
+        fi;
+    od;
+    
+    gensCentraliser := RECOG.InvolutionCentraliser(G,h,100);
+    # TODO: Compute generating set for OmegaX(E) (see paper from LGO)
+
+    #TODO: CONTINUE HERE
```
fingolfin, line 379: OK clearly this was never finished. So I wonder, how does the code handle $SL(4,q)$ right now?

After looking around a bit, it seems it uses `RECOG.SLn_godownfromd` for that -- which is written to work for any starting dimension `d`, but we only ever use it for `d=4`... Huh.

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+
+
+# Find random element s = r^PseudRandom(g) such that <r,s> is isomorphic to SL(4,q)
+# and check whether they are isomorphic
+RECOG.SLn_constructsl4:=function(g,dim,q,r)
+  local s,h,count,readydim4,readydim3,ready,u,orderu,
+        nullr,nulls,nullspacer,nullspaces,int,intbasis,nullintbasis,
+        newu,newbasis,newbasisinv,newr,news,outputu,mat,i,shorts,shortr;
+  nullr:=RECOG.FixspaceMat(r);
+  nullspacer:=VectorSpace(GF(q),nullr);
```
fingolfin, line 37: Might as well do it in one line, without a confusing extra variable...
```suggestion
  nullspacer:=VectorSpace(GF(q),RECOG.FixspaceMat(r));
```
... but I also wonder: why is this called `nullspacer` and not `fixspace_r` ? So like this (except that of course for such a change, all other references to `nullspacer` need to be adjusted)
```suggestion
  fixspace_r:=VectorSpace(GF(q),RECOG.FixspaceMat(r));
```

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+  repeat
+    s:=r^PseudoRandom(g);
+    nulls:=RECOG.FixspaceMat(s);
+    nullspaces:=VectorSpace(GF(q),nulls);
+    int:=Intersection(nullspacer,nullspaces);
+    intbasis:=Basis(int);
+    newbasis:=[];
+    for i in [1..Length(intbasis)] do
+        Add(newbasis,intbasis[i]);
+    od;
```
fingolfin, line 49:
```suggestion
    newbasis:=List(Basis(int));
```

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+    i:=0;
+    repeat
+       i:=i+1;
+       if not mat[i] in int then
+          Add(newbasis,mat[i]);
+          int:=VectorSpace(GF(q),newbasis);
+       fi;
+    until Dimension(int)=dim;
```
fingolfin, line 57: While this works, it is inherently wasteful, as the call to `\in` does some of the work which `VectorSpace` then has to do again (plus `VectorSpace` can't tell that the start of `newbasis` already forms a basis).

A better implementation might be to use `MutableBasis` together with `CloseMutableBasis` here.


Perhaps also `SumIntersectionMat` could be used to simplify the entire section (i.e. starting already in line 41)

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+        nullr,nulls,nullspacer,nullspaces,int,intbasis,nullintbasis,
+        newu,newbasis,newbasisinv,newr,news,outputu,mat,i,shorts,shortr;
+  nullr:=RECOG.FixspaceMat(r);
+  nullspacer:=VectorSpace(GF(q),nullr);
+  mat:=One(r);
+  ready:=false;
+  repeat
+    s:=r^PseudoRandom(g);
+    nulls:=RECOG.FixspaceMat(s);
+    nullspaces:=VectorSpace(GF(q),nulls);
```
fingolfin, line 43: Same questions/suggestions as above

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+    news:=newbasis*s*newbasisinv;
+
+    #shortr, shorts do not need memory
+    #we shall throw away the computations in h
+    #check that we have SL(4,q), by non-constructive recognition
+    
+    # Remark D.R.: Tries to reduce matrix multiplications
+    #               by working with 4 dimensional matrices
+    shortr:=newr{[dim-3..dim]}{[dim-3..dim]};
+    shorts:=news{[dim-3..dim]}{[dim-3..dim]};
```
fingolfin, line 70: I assume this means we are looking at stingray matrices with the "head/body" at the bottom right (instead of top left)?

A comment should stat that, I guess.

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+          readydim4:=true;
+       elif Gcd(orderu,(q^2+q+1)/Gcd(3,q-1))>1 then
+          readydim3:=true;
+       fi;
+       if readydim4 = true and readydim3 = true then
+          ready:=true;
+          break;
+       fi;
+       count:=count+1;
+    until count=30;
```
fingolfin, line 88: What's the idea here? Is this meant to be a test ("naming") to verify we have $SL_4$ ? Why is that correct, though?

I also wonder if it is better / cheaper / faster than `RecogniseClassical(h,rec(case:="linear"));

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+#############################################################################
+
+
+
+# TODO: Work on comments and documentation
+
+
+
+# Find random element s = r^PseudRandom(g) such that <r,s> is isomorphic to SL(4,q)
+# and check whether they are isomorphic
```
fingolfin, line 31: What is meant by "they"?  The group `<r,s>`, and... what?

gap/projective/constructive_recognition/SL/GoingDown.gi
```diff
@@ -0,0 +1,691 @@
...
+  until ready=true;
+
+  return Group(r,s);
+end;
+
+
+
+#g=SL(d,q), given as a subgroup of SL(dim,q)
+#output: [SL(2,q), and a basis for the 2-dimensional subspace where it acts
+RECOG.SLn_godownfromd:=function(g,q,d,dim)
```
fingolfin, line 97: It seems this is only ever called with `d=4`?

Is `dim` really necessary as an argument, can't we get that from `DimensionOfMatrixGroup(g)`?

Likewise for `q`, as `Size(FieldOfMatrixGroup(g))`

gap/projective/constructive_recognition/SL/GoingUp.gi
```diff
@@ -0,0 +1,533 @@
...
+        od;
+        
+        # Now cleanup column n above row n:
+        for j in [1..w.n-1] do
+            tf := DoColOp_n(tf,j,w.n,ci[j,w.n],w);
+        od;
+        
+        # Now cleanup row n left of column n:
+        for j in [1..w.n-1] do
+            tf := DoRowOp_n(tf,w.n,j,-c[i,j],w);
```
fingolfin, line 423: It seems this matches the paper draft @aniemeyer are working on, and disagrees with line 11 of Algorithm 26 in the thesis. Phew!

makedoc.g
```diff
@@ -21,7 +21,7 @@ if fail = LoadPackage("AutoDoc", ">= 2019.07.03") then
     ErrorNoReturn("AutoDoc 2019.07.03 or newer is required");
 fi;
 
-Read("regen_doc.g");
+Read(Filename(DirectoryCurrent(), "regen_doc.g"));
```
SoongNoonien: I think that this change is unnecessary.

regen_doc.g
```diff
@@ -109,9 +110,10 @@ local xmlfile, meth;
 end;
 
 GenerateMethodsListXML := function(shortname, db)
-    local xmlfile, dbsWhichUseMethod, nrDbsWhichUseMethod, s, meth;
+    local dir, xmlfile, dbsWhichUseMethod, nrDbsWhichUseMethod, s, meth;
 
-    xmlfile := Concatenation("doc/_methods_", shortname, "_list.xml");
+    dir := Filename(DirectoryCurrent(), "");
+    xmlfile := Concatenation(dir, "doc/_methods_", shortname, "_list.xml");
```
SoongNoonien: I think all these changes can be dropped as well. There shouldn't be the need to modify `regen_doc.g` at all.

fingolfin:
Superseded by #524, which is this code rebuilt on top of #523 (a pure,
behaviour-neutral split of `gap/projective/sl.gi` into the layout this branch
already uses). Leaving this open for reference; nothing here is lost.

Two things found while rebuilding that are worth recording here:

**This branch predates #506 and #516.** Copying its files verbatim silently
reverts the `SLn_godownfromd` eigenspace dimension check and drops the
`StripMemory` calls. #524 three-way merges each file instead.

**`RECOG.FindStdGens_SL` here still calls master's `RECOG.RecogniseSL2Natural`.**
So `SL/BaseCase.gi` and `SL/sl2_BlackBox.gi` — 1049 lines — are unreachable on
this branch, as are the whole `*SmallerMatrices*` family,
`FindStdGensSmallerMatrices_SL`, and `SL_godownToDimension2WithInvolutions`,
which calls `RECOG.CentraliserOfInvolution` and `RECOG.CompletionCheck` — names
defined nowhere on this branch. Of the 50 functions in `utils/utils.gi`, seven
are used here and 35 are called by nothing at all, even once #372's Sp/SU/O code
is present. `RECOG.EigenspaceMat` and `RECOG.FixspaceMat` there are
byte-identical duplicates of `gap/utils.gi` that load later and silently
override it, and `RECOG.InvolutionCentraliser` in `SL/BaseCase.gi` shadows the
four-argument one in `gap/projective/d247.gi`.

#524 keeps only what is reachable and moves the rest to #525, which addresses
the coverage item on this PR's TODO list. The old-vs-new `SLn_UpStep` diff,
921 lines of which 732 are indentation, is split there into a whitespace-only
commit and the 187 lines that actually change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #381, opened 2026-02-10
`RecogniseAn` recognises symmetric group as alternating group


TWiedemann:
This is ok:
```
gap> G := AlternatingGroup(9);
Alt( [ 1 .. 9 ] )
gap> RECOG.RecogniseAn(MovedPoints(G), G, 1/2000);
rec( conjperm := (1,7,8,2,4,9)(3,6,5), degree := 9, gens := [ (5,9,8), (1,7,4,5,2,6,3) ], stamp := "An" )
```
This is surprising:
```
gap> G := SymmetricGroup(9);
Sym( [ 1 .. 9 ] )
gap> RECOG.RecogniseAn(MovedPoints(G), G, 1/2000);
rec( conjperm := (1,7,6,9,2,4,8,3), degree := 9, gens := [ (3,9,8), (1,4,6,8,2,5,7) ], stamp := "An" )
```
However, this is reasonable in the broader scope: In order to decide whether `RecogniseAn` or `RecogniseSn` should be used, the function `RecogniseGiant` first checks whether all generators of its input group have sign 1. Hence `RecogniseGiant` will never call `RecogniseAn` with a full symmetric group. Thus the "error" above arises only if one directly calls the internal function `RecogniseAn`.

Suggested solution:
- Either document properly what `RecogniseAn` does and doesn't do,
- or get rid of `RecogniseSn` and `RecogniseAn` altogether and move their code directly into `RecogniseGiant`. I think I prefer this approach over having a `RecogniseAn` function which does not, in fact, recognise An.

The underlying reason for this behaviour is that `NiceGeneratorsAnOdd` and `NiceGeneratorsAnEven` only search for the nice generators of A_n in its input group, but do not check whether they actually generate the input group. I think this is reasonable, but again, it should be clarified in the documentation (which I will do in an upcoming PR).

fingolfin:
All good points. I have no strong opinion on this; if you think it'd be nicer to merge things together in `RecogniseGiant`, that'd be fine by me. But just clarifying the docs to note assumptions about the input would also be fine. Actually, checking the signs would be fine, too, it's not *that* expensive...

PR #382, opened 2026-02-10
`NiceGeneratorsAn` (`Odd` and `Even`): Improve documentation, rename variables, restructure code, add tests

TWiedemann:
The same changes as in #369 but for `NiceGeneratorsAnOdd`.

In addition, the random search for cycles of lengths 2, 3, and n in `NiceGeneratorsAnOdd` and `NiceGeneratorsSn` is moved to a new function `RECOG.FindCycles`.

Issue #381 is *not* resolved by this PR, but it is slightly related.

Update 19.02.2026: Added the same changes for `NiceGeneratorsAnEven` and updated `RECOG.FindCycles` to also accept n-1.

### DIFF
--- a/tst/working/quick/PermNiceGens.tst
+++ b/tst/working/quick/PermNiceGens.tst
@@ -58,35 +58,43 @@ gap> testNiceGeneratorsAn := function(grp)
 >    niceGens := RECOG.NiceGeneratorsAnEven(grp, 10000);
 >  fi;
 >  if niceGens <> fail then
->    longPerm := niceGens[1];
->    shortPerm := niceGens[2];
+>    # Up to renaming points, the following permutations should be...
+>    longPerm := niceGens[1]; # =(1,2)(3,...,m) for even m and =(3,...,m) for odd m
+>    shortPerm := niceGens[2]; # =(1,2,3)
 >    cyclenShort := CycleStructurePerm(shortPerm);
 >    cyclenLong := CycleStructurePerm(longPerm);
+>    if ForAny(Union([1], [3..m]), i->IsBound(cyclenShort[i])) or cyclenShort[2] <> 1 then
+>      Display("Short permutation is not a 3-cycle for m odd");
+>    fi;
 >    if IsOddInt(m) then
->      if ForAny(Union([1], [3..m]), i->IsBound(cyclenShort[i])) or cyclenShort[2] <> 1 then
->        Display("Short permutation is not a 3-cycle for m odd");
->      fi;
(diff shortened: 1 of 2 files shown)


gap/perm/giant.gi
```diff
@@ -20,6 +20,73 @@
...
 
+#########################################################################
+## For a permutation group grp on n points and a list lenList of integers,
+## returns either a list cycList such that cycList[i] is a cycle of length lenList[i]
+## for all 1 <= i <= Length(lenList), or fail.
+## Each entry of lenList must be either a prime number or n.
+## The desired circles are found by a random search, drawing at most N random
+## elements from grp.
+## Thus if grp actually contains cycles of the desired shape, then the function does not
+## return fail with a probability that grows as N grows.
```
fingolfin: Let's avoid double negation, e.g.
```suggestion
## Thus if grp actually contains cycles of the desired shape, then the function
## succeeds with a probability that grows as N grows.
```

tst/working/quick/PermNiceGens.tst
```diff
@@ -46,5 +48,46 @@ gap> for G in grps do if RECOG.NiceGeneratorsSn(G, 10000) <> fail then
...
+>      fi;
+>      if ForAny([1..m-4], i->IsBound(cyclenLong[i])) or cyclenLong[m-3] <> 1 then
+>        Display("Long permutation is not an (m-2)-cycle for m odd");
+>      fi;
+>      if Size(Intersection(MovedPoints(longPerm), MovedPoints(shortPerm))) <> 1 then
+>        Display("Supports of 3-cycle and (m-2)-cycle do not intersect in one point for m odd");
+>      fi;
+>    fi;
+
+# Test for even m is still missing
```
fingolfin: Please add `TODO:` or `FIXME:` to such comments, so that one can find them by grepping for those terms.
```suggestion
# TODO: Test for even m is still missing
```

TWiedemann:
I changed the test `"PSL(2,23)" in ri.possibleNearlySimple` as discussed. The CI check still fails, but I cannot reproduce this failure locally. Instead, I get an error in another line, but only on the first run of the test:
```
gap> Test("MoreNaming.tst");
########> Diff in MoreNaming.tst:941
# Input is:
ri := RecogniseClassical(maxes_cgo_plus_8_5[1]);;
# Expected output:
# But found:
#I  Have 38644 points.
########
false
gap> Test("MoreNaming.tst");
true
gap> Test("MoreNaming.tst");
true
gap> Test("MoreNaming.tst");
true
gap> Test("MoreNaming.tst");
true
gap> Test("MoreNaming.tst");
true
```

PR #391, opened 2026-03-03
Merge `RecogniseAn` and `RecogniseSn` into `RecogniseGiant`

### LINKED ISSUE #381: `RecogniseAn` recognises symmetric group as alternating group

TWiedemann:
- Resolves #381 by doing what the title says
- Similarly to #369 and #382, improves the documentation of `ConjEltSn`, `ConjEltAnEven` and `ConjEltAnOdd` and refactors the code slightly.

### DIFF
--- a/tst/working/quick/PermNiceGens.tst
+++ b/tst/working/quick/PermNiceGens.tst
@@ -2,8 +2,9 @@
 gap> START_TEST("PermNiceGens.tst");
 
 ####### Test NiceGeneratorsSn ########
+# (The tests for ConjEltSn make the tests before them obsolete, but there is no harm done.)
 gap> testNiceGeneratorsSn := function(grp)
->  local Sn, m, niceGens, fullCycle, transp, cyclen1, x, y, i;
+>  local Sn, m, niceGens, fullCycle, transp, cyclen1, x, y, i, conj;
 >  m := Size(MovedPoints(grp));
 >  niceGens := RECOG.NiceGeneratorsSn(grp, 10000);
 >  if m<=2 then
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
nice, thanks!

PR #394, opened 2026-03-13
Fix Frank's SL2 in nat.rep. code in even char

fingolfin:
Also add some Asserts for future debugging


frankluebeck:
Thanks for spotting this. The fix looks good to me.

Issue #338, opened 2025-04-28
Implement `StdPresentation`


fingolfin:
This is part of issue #48 but nowadays much closer at hand:

We now have the `clasicpres` package available, see <https://docs.gap-system.org/pkg/classicpres/htm/CHAP001.htm>, which we could use for this.

My suggestion would be to start small:
- implement presentations for a few leaf node cases (e.g. An/Sn -- or even cheat and use IsomorphismFpGroup for all permutation groups at the leaves... *for now*)
- add a generic method which lifts back presentations from images and kernels and stitches them together
- once these basics work we can cover more cases gradually

PR #402, opened 2026-03-25
Implement verification

### LINKED ISSUE #48: Needed: verification
Verification of the recognition results is completely missing.

For this, on the one hand we need (efficient) presentations for the recognized groups. On the other hand, we need to design and implement the infrastructure for recognition methods to specify those presentations; and then to effectively make use of them.

TWiedemann:
- Adds filter `IsVerified` for recog nodes and a documentation of its meaning.
- Adds a test to `RecogniseGeneric` that the nice generators actually generate the whole input group. This test is performed even if verification is not required, see the comments in the code.
- Implements the function `VerifyGroup` by calling `StdPresentation`.
- Ensures that `StdPresentation` works properly if it recognises an error (i.e. if it fails to rewrite a word).
- Removes the (unimplemented) functions `VerifyPermGroup`, `VerifyMatrixGroup`, `VerifyProjectiveGroup`. It is not clear to me why there would be separate functions for each type because `VerifyGroup` works fine for all types. (If some other function notices on the side that a recog node `ri` must be correct, then this function is free to set `IsVerified` for `ri`, but this does not require specific verification functions for each type.)

This is a draft because I have not yet written tests. It is likely that I will change something along the way, so no need to review yet.

This resolves #48 but not #338 because we still need to implement good presentations for the leaf nodes. (Actually #48 also mentions the presentations for the leaf nodes, so it is not as clear which one should be closed. In any case, there is clearly a big overlap. I can also create a new issue with the current status as soon as this PR is merged, and then we can close the other two.)

Uncertainties:
- The names `IsVerified` and `VerifyGroup` are up to debate, though the latter was already in the code.
- I am not sure what exactly `RecogniseGeneric` should do if it notices (before verification is done!) that something went wrong. Some earlier parts of `RecogniseGeneric` return `ri` without setting `IsReady` in this case, others return `fail`. https://github.com/gap-packages/recog/pull/402/changes#diff-7924d4adc61aac5f5af4083b894fec27355609be34562c7a425ac03aad8d5436R687-R698
- If `VerifyGroup` realises that something is wrong, it returns `fail` and the `IsReady` filter is removed by `CalcStdPresentation`. Maybe something else or something in addition should be done.
- ~~I haven't tried to build the new documentation.~~ Done.


TWiedemann:
This PR is now ready for review.
I recall:
> Uncertainties:
> 
>     * The names `IsVerified` and `VerifyGroup` are up to debate, though the latter was already in the code.
> 
>     * I am not sure what exactly `RecogniseGeneric` should do if it notices (before verification is done!) that something went wrong. Some earlier parts of `RecogniseGeneric` return `ri` without setting `IsReady` in this case, others return `fail`. https://github.com/gap-packages/recog/pull/402/changes#diff-7924d4adc61aac5f5af4083b894fec27355609be34562c7a425ac03aad8d5436R687-R698
> 
>     * If `VerifyGroup` realises that something is wrong, it returns `fail` and the `IsReady` filter is removed by `CalcStdPresentation`. Maybe something else or something in addition should be done.

gap/base/recognition.gi
```diff
@@ -815,19 +801,59 @@ InstallGlobalFunction( SLPforElementGeneric,
...
+        k,slp,relKer,x0,x,conj,presTot,
+        presAbort;
     # Child nodes
     riKer := KernelRecogNode(ri); # might be fail
     riFac := ImageRecogNode(ri);
     # If riKer = fail, then Homom(ri) is an isomorphism (or, in the projective case,
-    # induces an isomorphism between the central quotients) and hence there is nothing
-    # to do: We just take the presentation from riFac.
+    # induces an isomorphism between the central quotients) and hence there is
+    # nothing to do: We just take the presentation from riFac.
```
SoongNoonien, line 872: Is there a reason for this change?

gap/base/recognition.gi
```diff
@@ -892,8 +937,38 @@ InstallGlobalFunction( CalcStdPresentationGeneric,
...
+        fi;
+    od;
+    # StdPresentation will call CalcStdPresentation if no presentation has been computed
+    # so far.
+    if StdPresentation(ri) = fail then
+        return false;
+    else
+        SetFilterObj(ri, IsVerified);
+        return true;
+    fi;
```
SoongNoonien:
```suggestion
    fi;
    SetFilterObj(ri, IsVerified);
    return true;
```
Here we could reduce the indentation a bit.

gap/base/recognition.gi
TWiedemann: Yes, the comment line starting with "to do" looked like a "we still have to implement the following:..." to me, so I rephrased it.

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
...
+## this subtree, <M>H</M> is the group
+## associated to <A>ri</A> and <A>riKer</A> is the left child of <A>ri</A>,
+## then the group associated to <A>riKer</A> (or rather, its embedding into <M>H</M>)
+## equals the kernel of the homomorphism from <M>H</M> to the group associated
+## to the right child of <A>ri</A>.
+## Without verification, the group of <A>riKer</A> may be smaller than the kernel.
+## </Description>
+## </ManSection>
+## <#/GAPDoc>
+DeclareAttribute( "IsCorrect", IsRecogNode, "mutable" );
```
fingolfin:
```suggestion
DeclareProperty( "IsCorrect", IsRecogNode );
```

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
 ## <#/GAPDoc>
 DeclareAttribute( "CalcStdPresentation", IsRecogNode, "mutable" );
 
+## <#GAPDoc Label="IsCorrect">
+## <ManSection>
+## <Attr Name="IsCorrect" Type="Flag"/>
```
fingolfin:
```suggestion
## <Prop Name="IsCorrect"/>
```

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
...
 DeclareAttribute( "CalcStdPresentation", IsRecogNode, "mutable" );
 
+## <#GAPDoc Label="IsCorrect">
+## <ManSection>
+## <Prop Name="IsCorrect" Arg="ri"/>
+## <Description>
+## The recognition procedure may with a small probability produce recognition trees
+## that are not correct.
+## For this reason, it is possible to verify that a recognition tree
+## is mathematically correct by calling <A>IsCorrect</A>.
```
fingolfin:
```suggestion
## is mathematically correct by calling <C>IsCorrect</C>.
```
`A` is for *arguments :-)

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
...
+## However, this is in general expensive.
+## <P/>
+## A value <A>true</A> of <A>IsCorrect(ri)</A> signifies that the subtree rooted at <A>ri</A>
```
fingolfin:
```suggestion
## A value <K>true</K> of <C>IsCorrect(<A>ri</A>)</C> signifies that the subtree rooted at <A>ri</A>
```
(Not sure if the nesting of `A` inside `C` works here

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
...
+## has been verified to be correct.
+## A value of <A>false</A> signifies that there was an attempt of verification,
```
fingolfin:
```suggestion
## A value of <K>false</K> signifies that there was an attempt of verification,
```
K for keywords

gap/base/recognition.gd
```diff
@@ -231,6 +231,34 @@ DeclareAttribute( "StdPresentation", IsRecogNode, "mutable" );
...
+## but that this verification resulted in a proof that the subtree rooted at <A>ri</A>
+## is incorrect.
+## <P/>
+## Here correctness of a subtree means the following: If <A>ri2</A> is a node in
```
fingolfin:
```suggestion
## Here correctness of a subtree means the following: If <C>ri2</C> is a node in
```
and so on

gap/base/recognition.gi
```diff
@@ -457,12 +457,38 @@ InstallMethod( StdPresentation, [ IsRecogNode and HasCalcStdPresentation],
...
   end );
 
+InstallMethod( IsCorrect, [ IsRecogNode ],
+  function(ri)
+    local h;
+    # If verification was already done (possibly by other means than constructing
+    # a presentation)
+    if HasIsCorrect(ri) then
+        return IsCorrect(ri);
+    fi;
```
fingolfin: This check can go now: GAP stores the property, and ensures it is only computed once
```suggestion
```

gap/base/recognition.gi
```diff
@@ -457,12 +457,38 @@ InstallMethod( StdPresentation, [ IsRecogNode and HasCalcStdPresentation],
...
+        fi;
+    od;
+    # StdPresentation will call CalcStdPresentation if no presentation has been computed
+    # so far.
+    if StdPresentation(ri) = fail then
+        SetIsCorrect(ri, false);
+    else
+        SetIsCorrect(ri, true);
+    fi;
+    return IsCorrect(ri);
```
```suggestion
    return StdPresentation(ri) <> fail;
```

gap/base/recognition.gd
TWiedemann: I will try it out

TWiedemann: It works

gap/base/recognition.gi
```diff
@@ -708,9 +690,29 @@ InstallGlobalFunction( RecogniseGeneric,
...
+    for h in GeneratorsOfGroup( H ) do
+        if SLPforElement(ri, h) = fail then
+            # NiceGens(ri) is too small, abort
+            if InfoLevel(InfoRecog) = 1 and depth = 0 then Print("\n"); fi;
+            Info(
+                InfoRecog, 1,
+                "RecogNode <ri> could not be recognised,",
+                " IsReady(<ri>) is not set, recognition aborts"
+            );
+            return ri;
```
TWiedemann, line 712: Is this the correct way to abort recognition in `RecogniseGeneric`? (Note that this is done before any verification may take place.) I copied these lines from further above (lines 604-608 right now). However, in lines 582-585, instead `fail` is returned to abort the recognition process, and I do not see why the two situations are treated differently.

gap/base/recognition.gi
```diff
@@ -457,12 +457,28 @@ InstallMethod( StdPresentation, [ IsRecogNode and HasCalcStdPresentation],
...
+    for h in GeneratorsOfGroup( Grp(ri) ) do
+        if SLPforElement(ri, h) = fail then
+            return false;
+        fi;
+    od;
+    # StdPresentation will call CalcStdPresentation if no presentation has been computed
+    # so far.
+    return StdPresentation(ri) <> fail;
+end);
+
```
TWiedemann, line 475: Should this method also do something else to the recog node if it realises that something is wrong, or are we happy with simply setting `IsCorrect` to `false`? Maybe send a message to the user with `Info`?

TWiedemann:
I added two comments about potential improvements above, and which we also discussed on Tuesday (but I think without a final decision?).

gap/base/recognition.gi
fingolfin: This is fine as-is, no message should be generated. If the user invoked this function manually, they don't need an extra message.

However, other code that invokes `IsCorrect` for some reason should potentially  print one (as then the user did not invoke `IsCorrect` and so may not be aware that this could have happened)

fingolfin: I.e., this function is fine as it is :-)

gap/base/recognition.gi
fingolfin: The check which returns `fail` is much newer, from PR #139 (commit dd6865f58f0e0b0a34dccb092925d574cd3fb3bd) made in 2019. It used to be an `Assert` before that.

So the `return ri` pattern here is right. The code which does the `return fail` should be changed to `return ri` and also to produce a suitable `Info` message.

But that's not a problem of this PR. I'll open an issue about it.

fingolfin (approved):

PR #434, opened 2026-03-31
Fixes for RecogNaturalSL2

frankluebeck:
- mention and catch that algorithm only works for q >= 5
- simplified (and fixed?) search condition for ym
- made search for cm less restrictive (should now also work for q=5)
- fixed that in char two the expoment i was sometimes wrong


fingolfin:
Thanks we applied the fixes.

PR #435, opened 2026-04-07
Set name for functions in `BindRecogMethod` and document their origin

### LINKED ISSUE #430: Enhance `regen_doc.g` and `BindRecogMethod` to provide source locations
For developers it would be helpful if the manual chapter on recognition methods would also indicate where each method is defined (in which file and line). This could probably be achieved by enhancing `BindRecogMethod` (copying the method used by e.g. `InstallMethod` for this). That is, by using `INPUT_FILENAME()` and `INPUT_LINENUMBER()` to record the file and line being processed by GAP right now (these will indicate the place `BindRecogMethod` is being called from).

In addition, it should perhaps also call `SetNameFunction` on those functions. Right now they are anonymous, which is annoying for e.g. profiling. To wit:
```
gap> FindHomMethodsMatrix.BlockLowerTriangular;
<RecogMethod "BlockLowerTriangular": function( ri ) ... end>
gap> FindHomMethodsMatrix.BlockLowerTriangular!.func;
function( ri ) ... end
gap> NameFunction(last);
"unknown"
```
The function should be named `FindHomMethodsMatrix.BlockLowerTriangular`..

Only half that name is given half to `BindRecogMethod`:
```
BindRecogMethod(FindHomMethodsMatrix, "BlockLowerTriangular",
```
The problem is that the record `FindHomMethodsMatrix` does not know its name. Two ways to address that come to mind:
1. we only have a few such "databases" that occur as first argument, and we could just compare against each to figure out the name (i.e. "if first_arg = FindHomMethodsMatrix then name := "FindHomMethodsMatrix"; elif first_arg = ...`
   - easy hack, least invasive, but does not scale well to more "method databases"
2. we change the first argument of `BindRecogMethod` to be a string (so here `"FindHomMethodsMatrix"`), and use that; this string we can then turn into a variable via `ValueGlobal`
   - clean, will work in the future, but requires touching every single `BindRecogMethod` invocation
3. as a variant of 2, we could also let `BindRecogMethod` accept either a string or a record as first argument; this way one "gradually" migrate to the new calling syntax. However, I don't see the benefit for us (it would be useful if there was lots of external code using `BindRecogMethod`, but there isn't -- so we are better off changing everything at once)

@SoongNoonien perhaps you'd be willing to look into this next time?

SoongNoonien:
This implements part of #430.


fingolfin:
@SoongNoonien you wrote
> This implements part of https://github.com/gap-packages/recog/issues/430.

But which parts of #430 are missing?!

SoongNoonien:
Nothing is missing anymore :-) That was just my initial comment, in my second comment I mentioned that this PR now fixes #430 and I updated the title accordingly.

Issue #343, opened 2025-11-10
wrong `Size` for a `RecognizeGroup` result


ThomasBreuer:
With the following input, I get a too large `Size` for a `RecognizeGroup` result in almost half of the attempts.

```
G:= AtlasSubgroup( "B", 2 );;  # the group 2^(1+22).Co2
Ggens:= GeneratorsOfGroup( G );;

F:= GF(2);;
M:= GModuleByMats( Ggens, F );;
facts:= MTX.CompositionFactors( M );;
Length( facts );  # 27
S:= First( facts, x -> x.dimension = 22 );;
C:= GroupWithGenerators( S.generators );;  # isom. with Co2
nice:= NiceMonomorphism( C );;
permgens:= List( S.generators, x -> ImagesRepresentative( nice, x ) );;
memgens:= GeneratorsWithMemory( permgens );;
find:= AtlasProgram( "Co2", "find" );;
res:= ResultOfBBoxProgram( find.program, GroupWithGenerators( memgens ) );;
slps:= List( res, SLPOfElm );;
stdmats:= List( slps, x -> ResultOfStraightLineProgram( x, Ggens ) );;

mx:= List( Maxes( CharacterTable( "Co2") ), CharacterTable );;
pos:= PositionsProperty( mx, x -> Size( x ) mod 23 = 0 );  # [ 9 ]
mx{ pos };  # [ CharacterTable( "M23" ) ]
restprg:= AtlasProgram( "Co2", "maxes", pos[1], 1 );;  # std. gens. of M23
stdrestmats:= ResultOfStraightLineProgram( restprg.program, stdmats );;
U:= GroupWithGenerators( stdrestmats );;  # 2^(1+22).M23 (or a subgroup)

M2:= GModuleByMats( stdrestmats, F );;
bases:= MTX.BasesCompositionSeries( M2 );;
Length( bases[7] );  # 198

ind:= MTX.InducedActionSubmodule( M2, bases[7] );;
U:= GroupWithGenerators( ind.generators );;

LoadPackage( "recog" );
Size( RecognizeMatrixGroup( U ) ) = 2^23 * Size( MathieuGroup( 23 ) );  # true
# So the 198x198 matrices really generate 2^(1+22).M23

mats:= ind.generators;;
mx:= AtlasProgram( "M23", "maxes", 7 );  # 23:11 in M23
res:= ResultOfStraightLineProgram( mx.program, ind.generators );;

sizes:= [];
[...]

PR #446, opened 2026-04-09
Add std presentation for LowerLeftPGroup

fingolfin:
This exploits that the nice generators in this case form a pcgs, and we have efficient methods to work with these.

This does *not* compute a pc group, to fit in the general framework, but perhaps we should add that at some point?

Also speed up CalcStdPresentationGeneric substantially by using GroupHomomorphismByImagesNC (otherwise it ends up trying to do a membership test for the range, which of course we can't do in general for our larger groups).

---

Follow-up to PR #441; motivated by issue #343 (but of course we want something like this anyway.


TWiedemann (approved):
Looks good to me.

PR #448, opened 2026-04-09
Fix C6 block kernel handling

fingolfin:
When Blocks sees blocksize 1, BlocksModScalars is trivial. Pass the whole group to BlocksBackToMats in that case.

Resolves #444.

UPDATE: modified, now also fixes #471.


fingolfin:
I've marked this as draft as I am not yet quite convinced this is the correct fix for the underlying issue, or just curing a symptom.

Issue #379, opened 2026-02-10
Classical naming: failures to name a classical group as what it is


fingolfin:
The test file `tst/working/combined/ClassicalNaturalNaming.tst` suggests that naming for certain classical groups is less reliable than in the general.

This may not be an actual "issue" but a normal artifact of these things being randomized; after all, in the "real world" application, we'll just retry the renaming eventually.

Still, I thought it would be good to settle this explicitly, hopefully with some feedback by @aniemeyer .

I've decided to turn this into an issue with concrete reproduces. To this end, I've been running this code for various parameters `type`, `d`, and `q` (requires PR #378, so make sure to `git pull` the latest `master`):
```gap
ReadPackage("recog", "tst/naming.g");
total:=0;;
for i in [1..100] do
  Reset(GlobalMersenneTwister,i);;
  Print("seed ",i,"\n");
  total := total + TestNaming(type,d,q);
od;
total;
```
or if you prefer a one-line for easier repetitions:
```gap
ReadPackage("recog", "tst/naming.g");;total:=0;;for i in [1..100] do Reset(GlobalMersenneTwister,i);; Print("seed ",i,"\n");total:=total+TestNaming(type,d,q);od;total;
```

Note that `TestNaming` invokes `RecogniseClassical` 20 times, so really we are testing it 20*100=2000 times. 

## Unitary groups

For $d=6$, $q=2$ we get 21 out of 2000 failures, but other parameters seem fine.

## Symplectic groups

For $d=4$ all is good for $q=2,3,5,7,8,11$ (though for 8 it is rather slow). But not so for $q=4,9,13$ (with 2000 (!!!), 790 (!), 101 failures out of 2000). Yes, it *always* fails for `Sp(4,4)`. It is always "has bad value for isSpContained; expected true, got unknown"

For $d=6$, $q=2$ we get 12 out of 2000 failures, but for larger $d$ and/or $q$ things seem to be fine.

## Orthogonal groups in odd dimension

For $d=3$ and $q=2,3,4,5,8,13,16,17$ all is good (for 20*100=2000 tests). For $q=7$ we see a few failures (just 3 out of 20*100 tests):
```gap-repl
gap> Reset(GlobalMersenneTwister,8);; TestNaming("SO", 3, 7);
1: SO(3,7) has bad value for isOmegaContained; expected true, got unknown
```

[...]

PR #451, opened 2026-04-14
Avoid naming Singer normalizers as SL

fingolfin:
Delay ruling out extension-field cases for prime-dimensional linear groups until the observed ppd data is incompatible with the Singer normalizer pattern.

Fixes #442

---

This was a "one shot" solution by the AI. I have not yet fully understood the `d > 3` change. I'll get backt to this when I have some more time.

I think a similar issue affects `IsGenericParameters`.

### DIFF
--- a/tst/naming.g
+++ b/tst/naming.g
@@ -4,11 +4,11 @@ TestNaming := function(grpname, param...)
     expected := rec(
         #isGeneric := true,
         isNotAlternating := true,
-        #isNotExt := true,
+        #isNotExt := true,  # FIXME: wrong value for SO(1,8,q)
         isNotMathieu := true,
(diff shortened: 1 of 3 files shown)


tst/naming.g
```diff
@@ -4,11 +4,11 @@ TestNaming := function(grpname, param...)
     expected := rec(
         #isGeneric := true,
         isNotAlternating := true,
-        #isNotExt := true,
+        #isNotExt := true,  # FIXME: wrong value for SO(1,8,q)
```
fingolfin, line 7: I'll deal with this in a follow-up PR

fingolfin:
Seem this also fixed a large chunk of issue #379

PR #458, opened 2026-04-21
Run An/Sn tests also in degrees 15, 20, 25 and more

fingolfin:
... as e.g. `RECOG.ConstructXiSn` and `RECOG.ConstructXiAn` have special cases for degrees divisible by 5 that only trigger the first time at these degrees.


fingolfin:
Turns out there is some code that require odd degree, multiple of 5, at leas 4 times 5... so we also need to include 25 as well. But the tests are so quick, I've just upped it to 30 (we should maybe move some of this from "slow" to "quick" anyway)

PR #461, opened 2026-04-24
suzuki example

fingolfin:
Co-authored-by: Alice Niemeyer <Alice.Niemeyer@art.rwth-aachen.de>


fingolfin:
The failing test in GAP devel is also on master.

PR #465, opened 2026-04-28
Add good `StdPresentation` for symmetric and alternating groups

TWiedemann:
Minor changes beside the title:
- Fixes `TryFindHomMethod`. (From what I understand, this adapts `TryFindHomMethod` to a change that was made in a recent PR, maybe someone knows which one.)
- Expands the documentation of the recog method "MovesOnlySmallPoints".
- Clarifies that the recog method "Giant" works only for permutation groups of degree at least 8 because `IsGiant` does.

### DIFF
--- a/doc/recog.bib
+++ b/doc/recog.bib
@@ -296,6 +296,21 @@ MRREVIEWER = {P. P. Pálfy},
       NOTE = {https://doi.org/10.1080/00927879508825509},
 }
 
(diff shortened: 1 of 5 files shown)


TWiedemann:
The documentation now builds correctly, only the hash key error remains.

fingolfin (approved):

PR #470, opened 2026-05-07
Don't return 'fail' in RecogniseGeneric

fingolfin:


TWiedemann (approved):
Changes look good, CI error seems unrelated, CodeCov complaint seems tolerable. I remain clueless about what the line breaks printed in lines 567 and 610 (which were not introduced in this PR) achieve, however.

Issue #166, opened 2019-10-28
Perhaps add back black box interface? In particular, for groups GL(d,Z/mZ).


fingolfin:
I removed this not so long ago, because (a) nothing was implemented, and (b) no good uses cases were brought forward by anybody. But now for (b), it was pointed out to me that inputs in GL_d(R), with R=Z/mZ or R a field in char 0, could be of interested (in the latter case one of may should presuppose that the input group is known to be finite). In those cases, we can compute products and inverses, and test for equality effectively. Also computing orders should be reasonably effective.

Of course to be actually useful, somebody would need to implement some of the basic algorithms, such as An/Sn recognition based on Jambor et al, SL_2, etc. 

For inputs in GL_d(R), one should also be able to reduce to SL_d(R) with the usual methods; and for R=Z/mZ, one can use that this ring splits into a sum of rings Z/q_iZ with q_i prime powers; from there, which induces an effective isomorphism $GL(d,Z/mZ) \cong GL(d,Z/q_1Z) x ... x GL(d,Z/q_kZ)$, so one can reduce to GL(d,Z/qZ); from this one gets an effective epi onto GL(d,p)$, where p is the unique prime dividing q. So I *think* that should make it possible to work rather well with these types of groups.

PR #485, opened 2026-05-19
Restore compatibility with the matgrp package

fingolfin:
... via a crude hack.

Thing is, the arguments for BindRecogMethod, as well as the inputs and outputs of recognition methods have changed.

Luckily the method in matgrp does nothing very useful, so we can replicate an updated version here.

---

See also https://github.com/hulpke/matgrp/pull/19 which addresses this on the matgrp side by simply removing the offending code there (it doesn't do anything useful as far as I can tell). But since we need to be compatible with the current release of matgrp, too, I am making this hackish workaround here (I've also thought about doing a "proper" adapter that is more general, but I don't know of any other external package using these APIs, so I opted to keep it simple.

### DIFF
--- a/gap/base/methods.gi
+++ b/gap/base/methods.gi
@@ -86,7 +86,22 @@ end);
 InstallGlobalFunction(BindRecogMethod,
 function(rname, arg...)
     local r, name;
-    r := ValueGlobal(rname);
+    if IsString(rname) then
+      r := ValueGlobal(rname);
+    elif rname = FindHomMethodsMatrix and arg[1] = "Nonfield" and EndsWith(INPUT_FILENAME(), "recograt.gi") then
+      # FIXME/HACK: backwards compatibility for the matgrp package
+      rname := "FindHomMethodsMatrix";
+      r := FindHomMethodsMatrix;
+      arg[Length(arg)] :=
+            function(ri)
+              if IsBound(ri!.ring) and not IsBound(ri!.field) then
+                Error("hereIAm");
+              fi;
+              return NeverApplicable;
+            end;
+    else
+      Error("<rname> must be the name of a global variable containing a record");
+    fi;
     r.(arg[1]) := CallFuncList(RecogMethod, arg);
     name := Concatenation(rname, ".", arg[1]);
     SetNameFunction(r.(arg[1])!.func, name);


gap/base/methods.gi
```diff
@@ -86,7 +86,22 @@ end);
...
+    if IsString(rname) then
+      r := ValueGlobal(rname);
+    elif rname = FindHomMethodsMatrix and arg[1] = "Nonfield" and EndsWith(INPUT_FILENAME(), "recograt.gi") then
+      # FIXME/HACK: backwards compatibility for the matgrp package
+      rname := "FindHomMethodsMatrix";
+      r := FindHomMethodsMatrix;
+      arg[Length(arg)] :=
+            function(ri)
+              if IsBound(ri!.ring) and not IsBound(ri!.field) then
+                Error("hereIAm");
```
SoongNoonien, line 98: What does this error mean? I see that it has been copied from matgrp but I'm not sure what it means.

fingolfin: I think it means that this is a debug left-over that should just be removed from `matgrp` ;-)

Perhaps @hulpke was working on support for $\mathrm{GL}(d,\mathbb{Z}/m\mathbb{Z})$ as input to `RecognizeGroup` (see also #166), and this was meant to ensure nothing of that sort crept into the wrong places. But this is just speculation.

gap/base/methods.gi
```diff
@@ -86,7 +86,22 @@ end);
...
 function(rname, arg...)
     local r, name;
-    r := ValueGlobal(rname);
+    if IsString(rname) then
+      r := ValueGlobal(rname);
+    elif rname = FindHomMethodsMatrix and arg[1] = "Nonfield" and EndsWith(INPUT_FILENAME(), "recograt.gi") then
+      # FIXME/HACK: backwards compatibility for the matgrp package
+      rname := "FindHomMethodsMatrix";
+      r := FindHomMethodsMatrix;
+      arg[Length(arg)] :=
```
SoongNoonien, line 95: I don't get this. I would have thought that the above two lines should be sufficient. By "above two" I mean 93 and 94.

fingolfin: But the number of arguments for recognition methods has dropped from 2 to 1; and the return values have switched from true/fail/false to string constants Success/TemporaryFailure/NeverApplicable. So we can't use the given function as-is, and must either wrap it (to convert inputs and outputs), or, as I chose here, replace it.

SoongNoonien: Ah, yes. Now I get it. I've overlooked the removal of `G`...

SoongNoonien (approved):

PR #487, opened 2026-05-26
Make `RECOG.IsThisSL2Natural` more precise

SoongNoonien:
This is a test to see if a change in #371 makes sense.

### DIFF
--- a/gap/projective/classicalnatural.gi
+++ b/gap/projective/classicalnatural.gi
@@ -854,6 +854,9 @@ function(ri)
   RECOG.SetPseudoRandomStamp(g,"ClassicalNatural");
 
   # First check whether we are applicable:
+  # check if this group contains SL_d -- for d > 2 we can use RecogniseClassical;
+  # for d = 2, we can use RECOG.IsThisSL2Natural, but only after adjusting
+  # the determinants of the generators (comes next)
   if d <> 2 then
       classical := RecogniseClassical(g);
       if classical.isSLContained <> true then
--- a/gap/projective/classicalnatural.gi
+++ b/gap/projective/classicalnatural.gi
@@ -854,12 +854,7 @@ function(ri)
   RECOG.SetPseudoRandomStamp(g,"ClassicalNatural");
 
   # First check whether we are applicable:
-  if d = 2 then
-      if not RECOG.IsThisSL2Natural(GeneratorsOfGroup(g),f) then
-          Info(InfoRecog,2,"ClassicalNatural: Is not PSL_2.");
-          return TemporaryFailure; # FIXME: TemporaryFailure here really correct?
-      fi;
-  else
+  if d <> 2 then
       classical := RecogniseClassical(g);
       if classical.isSLContained <> true then
           Info(InfoRecog,2,"ClassicalNatural: Is not PSL.");
@@ -884,6 +879,15 @@ function(ri)
           changed := true;
       fi;
   od;
+
+  # Now check whether the normalized matrices generate an SL_2:
+  if d = 2 then
+      if not RECOG.IsThisSL2Natural(gens,f) then
+          Info(InfoRecog,2,"ClassicalNatural: Is not PSL_2.");
+          return TemporaryFailure; # FIXME: TemporaryFailure here really correct?
+      fi;
+  fi;
+
   if changed then
       gm := GroupWithMemory(gens);
       pr := ProductReplacer(GeneratorsOfGroup(gm),rec(maxdepth := 500));
--- a/gap/projective/classicalnatural.gi
+++ b/gap/projective/classicalnatural.gi
@@ -654,7 +654,7 @@ RECOG.IsThisSL2Natural := function(gens,f)
       Info(InfoRecog,4,"SL2: Computing stabiliser chain.");
       S := StabilizerChain(Group(gens));
       Info(InfoRecog,4,"SL2: size is ",Size(S));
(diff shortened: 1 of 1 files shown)


gap/projective/classicalnatural.gi
```diff
@@ -654,7 +654,7 @@ RECOG.IsThisSL2Natural := function(gens,f)
       Info(InfoRecog,4,"SL2: Computing stabiliser chain.");
       S := StabilizerChain(Group(gens));
       Info(InfoRecog,4,"SL2: size is ",Size(S));
-      return Size(S) mod (q*(q-1)*(q+1)) = 0;
+      return Size(S) = (q*(q-1)*(q+1));
```
fingolfin, line 654: Thanks for starting this experiment.

So now we know: With this change we get a bunch of errors in tests related to the naming / recognition of $(P)SL(2,q)$:
```
testing: /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst
########> Diff in /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst:60
# Input is:
TestRecogGL(2,4);;
# Expected output:
Stamp: ClassicalNatural
# But found:
Stamp: LowIndex
########
########> Diff in /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst:86
# Input is:
TestRecogGL(2,5);;
# Expected output:
Stamp: ClassicalNatural
# But found:
Stamp: LowIndex
########
########> Diff in /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst:112
# Input is:
TestRecogGL(2,8);;
# Expected output:
Stamp: ClassicalNatural
# But found:
Stamp: StabilizerChainProj
########
########> Diff in /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst:138
# Input is:
TestRecogGL(2,9);;
# Expected output:
Stamp: ClassicalNatural
# But found:
Stamp: StabilizerChainProj
########
########> Diff in /tmp/gaproot/pkg/recog/tst/working/veryslow/ClassicalNatural.tst:154
# Input is:
TestRecogGL(10,9);;
# Expected output:
Stamp: ClassicalNatural
# But found:
ALARM: set count to -1 to skip test!
...
```
So at least with the current code, we need this.

But it may still be that it only papers over a symptom instead of fixing a root cause. Note that the code originally had the stricter `=` test and I weakened it in 7bd8018eb1ef7c2f0f51957744dcdbd84a9caaeb

Thing is, the name of this function strongly suggests it is meant to detect the matrix group $SL_2$; but this code and some other comments it is meant to detect $PSL_2$. This discrepency should be resolved, one way or the other.

fingolfin: Looking at the (currently) only caller for `RECOG.IsThisSL2Natural`, we see this:
```gap
  # First check whether we are applicable:
  if d = 2 then
      if not RECOG.IsThisSL2Natural(GeneratorsOfGroup(g),f) then
          Info(InfoRecog,2,"ClassicalNatural: Is not PSL_2.");
          return TemporaryFailure; # FIXME: TemporaryFailure here really correct?
      fi;
  else
      classical := RecogniseClassical(g);
      if classical.isSLContained <> true then
          Info(InfoRecog,2,"ClassicalNatural: Is not PSL.");
          return TemporaryFailure; # FIXME: TemporaryFailure here really correct?
      fi;
  fi;

  # Now get rid of nasty determinants:
  gens := ShallowCopy(GeneratorsOfGroup(g));
  changed := false;
  gcd := Gcdex(d,q-1);
  for i in [1..Length(gens)] do
      det := DeterminantMat(gens[i]);
      if not IsOne(det) then
          root := RootFFE(f, det^-1, d);
          if root = fail then
              Info(InfoRecog,2,
                   "ClassicalNatural: determinant cannot be normalized in field.");
              return fail;
          fi;
          gens[i] := gens[i] * root;
          changed := true;
      fi;
  od;
```

So: we first call `RECOG.IsThisSL2Natural` or `RecogniseClassical`, and only if those give a "green light' do we proceed to adjust the matrices to have "nice" determinant.

So one alternate fix (which might allow merging this PR here) could be to interchange these two code blocks, and "normalize" the matrices first.

(This PR should not be merged in its current form, though: at the very least, some of the code comments added in 7bd8018eb1ef7c2f0f51957744dcdbd84a9caaeb should be adjusted accordingly, or be removed again)

SoongNoonien: Ok, I tried to reorder these two blocks and we'll see if it works now.

gap/projective/classicalnatural.gi
```diff
@@ -854,12 +854,7 @@ function(ri)
   RECOG.SetPseudoRandomStamp(g,"ClassicalNatural");
 
   # First check whether we are applicable:
-  if d = 2 then
-      if not RECOG.IsThisSL2Natural(GeneratorsOfGroup(g),f) then
-          Info(InfoRecog,2,"ClassicalNatural: Is not PSL_2.");
-          return TemporaryFailure; # FIXME: TemporaryFailure here really correct?
-      fi;
-  else
+  if d <> 2 then
```
fingolfin, line 857:
```suggestion
  # check if this group contains SL_d -- for d > 2 we can use RecogniseClassical;
  # for d = 2, we can use RECOG.IsThisSL2Natural, but only after adjusting
  # the determinants of the generators (comes next)
  if d <> 2 then
```

fingolfin (approved):

PR #493, opened 2026-06-09
Fix docs github action

SoongNoonien:
Currently, all docs github actions fail even on master. This PR ignores warnings of which we have a lot due to the dynamic structure of the docs.


fingolfin (approved):
Maybe also open an issue about these warnings, so that one day we'll fix them?

PR #504, opened 2026-06-23
add RECOG.FindSL2inSL4: Find SL(2,q) as stingray subgroup of SL(4,q)

Till-Eisen:
This PR adds `gap/projective/SL4_natural.gi`, which implements `RECOG.FindSL2inSL4(G, N)`.

Given G = SL(4,q) (q odd), the function finds a subgroup U isomorphic to SL(2,q) stingray-embedded in G: concretely, a base change matrix `bas` is computed such that every element of U is block-diagonal diag(A, I_2) in the new basis.

### DIFF
--- a/gap/projective/SL4_natural.gi
+++ b/gap/projective/SL4_natural.gi
@@ -56,7 +56,7 @@
 #############################################################################
 
 
-RECOG.GoingDownFinalStepSL := function(G, N)
+RECOG.FindSL2inSL4 := function(G, N)
     local n, F, q, one, Eigenspace, gens_group, gens_bas, l,
           pr, t, basis1, basism1, bas, basInv,
           h, hb, topLeft, bottomRight, ordr, K, slp, Nstart, U;
--- a/gap/projective/SL4_natural.gi
+++ b/gap/projective/SL4_natural.gi
@@ -1,0 +1,147 @@
+###############################################################################
+##
+##  Authors: Till Eisenbrand
+##           The underlying algorithm is due to Daniel Rademacher; see
+##           Algorithm 15 ("GoingDownFinalStepSL") in:
+##           D. Rademacher, "Constructive Recognition of Finite Classical Groups with
+##           Stingray Elements", Ph.D. thesis, RWTH Aachen University, Germany, 2024.
+##
+##
+##  This file provides the function:
+##
+##    RECOG.GoingDownFinalStepSL(G, N)
+##
+##  Purpose:
+##    Given G = SL(4,q) (q odd), this function finds an embedded copy of
+##    SL(2,q) inside G as a "stingray subgroup".  Concretely, it looks for
+##    a base change matrix  bas  such that, in the new basis, the found
+##    subgroup consists of block matrices of the form
+##
+##        [ A | 0 ]
+##        [ 0 | I ]      with  A in SL(2,q).
+##
+##    The function proceeds by:
+##      1. Searching for a strong pre-involution  t  of type 2+2, i.e. a
+##         non-trivial involution with dim(E_1(t)) = dim(E_{-1}(t)) = 2.
+##      2. Building a base change from the eigenspaces of  t.
+##      3. Drawing random elements from the centraliser C_G(t) via the Bray
+##         trick (RECOG.CentralisingElementOfInvolution) and filtering for
+##         those that act trivially on the lower 2x2 block in the new basis.
+##         Such elements generate the stingray-embedded SL(2,q).
+##      4. Stopping as soon as the collected generators are recognised
+##         (non-constructively) as SL(2,q).
+##
+##  Input:
+##    G  -- a matrix group equal to SL(4,q) for some odd prime power q,
+##           given by 4x4 matrices over GF(q).
+##    N  -- positive integer: random-element budget (number of allowed
+##           calls to the product replacer / Bray trick).
+##
+##  Returns:
+##    "fail"  if the budget N is exhausted before success, OR a record with
+##    the following fields:
+##
+##      .U     -- a subgroup of G with U isomorphic to SL(2,q), stingray
+##                embedded in G with respect to the base change bas.
+##
+##      .bas   -- the 4x4 base change matrix bas such that  u^bas
+##                is block-diagonal diag(A, I_2) for every u in U.
+##
+##      .gens  -- list of generators of U as elements of G (i.e. 4x4
+##                matrices over GF(q)).
+##
+##      .Nout  -- remaining budget after completion  (N - number of random
+##                selections actually used).
+##
+#############################################################################
+
+
+RECOG.GoingDownFinalStepSL := function(G, N)
+    local n, F, q, one, Eigenspace, gens_group, gens_bas, l,
+          pr, t, basis1, basism1, bas, basInv,
+          h, hb, topLeft, bottomRight, ordr, K, slp, Nstart, U;
+
+    n := DimensionOfMatrixGroup(G);
+    F := FieldOfMatrixGroup(G);
+    q := Size(F);
+    one := IdentityMat(n, F);
+    Nstart := N;
+    slp := [];
+
+    if q <10 then
+        l := 3;
+    else 
+        l := 3;
+    fi;
+
+    if n <> 4 then
+        Error("GoingDownFinalStepSL: <G> must act in dimension 4");
+    fi;
+    if q mod 2 = 0 then
+        Error("GoingDownFinalStepSL: q must be odd");
+    fi;
+
+    Eigenspace := function(M, lambda)
+        return NullspaceMat(M - lambda * one);
+    end;
+
+    pr := ProductReplacer(GeneratorsOfGroup(G));
+
+    # find a strong pre-involution: t = x^(|x|/2) of type 2+2.
+    # RECOG.InvolutionSearcher(pr,ord,0) uses exactly one Next(pr).
+    t := fail;
+    while N > 0 do
+        t := RECOG.InvolutionSearcher(pr, Order, 0);
+        N := N - 1;
+        if t <> fail and Length(Eigenspace(t, One(F))) = 2 then
+            break;
+        fi;
+        t := fail;
+    od;
+
+    if t = fail then
+        Print("GoingDownFinalStepSL: out of budget while looking for a strong pre-involution\n");
+        return "fail";
+    fi;
+
+    basis1 := Eigenspace(t, One(F));
+    basism1 := Eigenspace(t, -One(F));
+    basInv := Concatenation(basis1, basism1);
+    bas := basInv^-1;
+
+    gens_bas := [];
+    gens_group := [];
+
+    while N > 0 do
+        h := RECOG.CentralisingElementOfInvolution(pr, Order, t);
+        N := N - 1;
+
+        # Conjugate h into the eigenbasis of t; in this basis C_G(t)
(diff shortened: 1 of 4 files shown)


gap/projective/sl4_natural.gi
fingolfin, line 1: This file should be called `sl4_natural.gi` (in all lowercase).

Also, you must actually load it (in `read.g`), like its SL2 counterpart

gap/projective/SL4_natural.gi
```diff
@@ -0,0 +1,147 @@
...
+    q := Size(F);
+    one := IdentityMat(n, F);
+    Nstart := N;
+    slp := [];
+
+    if q <10 then
+        l := 3;
+    else 
+        l := 3;
+    fi;
```
fingolfin: ?

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,142 @@
...
+    one := IdentityMat(n, F);
+    Nstart := N;
+    slp := [];
+
+    if q <10 then
+        l := 3;
+    else 
+        l := 3;
+    fi;
+
```
fingolfin:
```suggestion
    l := 3;

```

tst/working/quick/sl4.tst
```diff
@@ -0,0 +1,31 @@
...
+>     if hb{[1,2]}{[3,4]} <> NullMat(2, 2, F) then
+>       return false;
+>     fi;
+>     if hb{[3,4]}{[1,2]} <> NullMat(2, 2, F) then
+>       return false;
+>     fi;
+>   od;
+>   return true;
+> end;;
+gap> list := Filtered([3..100], q -> IsPrimePowerInt(q) and q mod 2 = 1);;
```
fingolfin: maybe more readable:
```suggestion
gap> list := Filtered([3..100], q -> IsOddInt(q) and IsPrimePowerInt(q));;
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,142 @@
...
+            Add(gens_bas, topLeft^ordr);
+            Add(gens_group, h^ordr);
+            
+            # Wait until we have enough generators before testing,
+            # to avoid calling RecognizeGroup on trivial/cyclic groups.
+            if Size(gens_bas) >l then
+                K := Group(gens_bas);
+                # Non-constructive recognition: check whether the collected
+                # 2x2 generators already span SL(2,q).
+                if Size(RecognizeGroup(K)) = Size(SL(2, q)) then
```
fingolfin:
```suggestion
                if Size(RecognizeGroup(K)) = SizeSL(2, q) then
```

fingolfin: This kind of check is OK for a quick experiment, but of course not for the final code: you throw away the recognition information which we then need to recover once more.

Instead I assume this should be directly invoking the SL_2 recognition code you add recently.

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,142 @@
...
+        Error("GoingDownFinalStepSL: <G> must act in dimension 4");
+    fi;
+    if q mod 2 = 0 then
+        Error("GoingDownFinalStepSL: q must be odd");
+    fi;
+
+    Eigenspace := function(M, lambda)
+        return NullspaceMat(M - lambda * one);
+    end;
+
```
fingolfin: I've just added `RECOG.FixspaceMat` and `RECOG.EigenspaceMat` to master via PR #506, so you can remove this helper and instead call those new ones
```suggestion
```

gap/projective/SL4_natural.gi
Till-Eisen: I was testing what minimal number of generators should be used until the verification process starts, i.e. what is the fastest running time. However, the verification should be done differently, as you already written in a comment. The number will depend on $q$ and the verification process. In this case $\ell=3$ is always optimal. However, with a different verification, $\ell = 2$ or maybe even $\ell =1$ for $q$ large enough, could be sufficient and thereby optimize the running time.

gap/projective/sl4_natural.gi
Till-Eisen: Here we need the non-constructive recognition of $SL(2,q)$. I am not certain about the proper function that should be used. 

Recently I added the constructive recognition of $SL(2,q)$.

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+    while N > 0 do
+        t := RECOG.InvolutionSearcher(pr, Order, 0);
+        N := N - 1;
+        if t <> fail and Length(RECOG.FixspaceMat(t)) = 2 then
+            break;
+        fi;
+        t := fail;
+    od;
+
+    if t = fail then
```
fingolfin: I'd test `N = 0` here because it could happen that `N = 0` yet `t <> fail` -- and in that case, we end up doing a couple more computations, only to return `fail` in the end anyway.

Plus, we can get rid of the `t := fail` in lines 85 and 78 
```suggestion
    od;

    if N = 0 then
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+        Print("FindSL2inSL4: out of budget while looking for a strong pre-involution\n");
```
fingolfin: This should be an Info message

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+    pr := ProductReplacer(GeneratorsOfGroup(G));
+
+    # find a strong pre-involution: t = x^(|x|/2) of type 2+2.
+    # RECOG.InvolutionSearcher(pr,ord,0) uses exactly one Next(pr).
+    t := fail;
+    while N > 0 do
+        t := RECOG.InvolutionSearcher(pr, Order, 0);
+        N := N - 1;
+        if t <> fail and Length(RECOG.FixspaceMat(t)) = 2 then
+            break;
```
fingolfin: It is very minor, but we can avoid computing the fixespace of `t` twice by storing it here:
```suggestion
        if t <> fail then
            basis1 := RECOG.FixspaceMat(t);
            if Length(basis1) = then
                break;
            fi;
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+        # Conjugate h into the eigenbasis of t; in this basis C_G(t)
+        # is block-diagonal diag(A, B) with A,B in GL(2,q).
+        hb := basInv * h * bas;
+        topLeft := hb{[1,2]}{[1,2]};
+        bottomRight := hb{[3,4]}{[3,4]};
+
+        # h^ordr kills the lower block (bottomRight^ordr = I_2) while
+        # potentially leaving a non-trivial upper block.
+        ordr := Order(bottomRight);
+        if topLeft^ordr <> IdentityMat(2, F) then
```
fingolfin: This creates an identity matrix and discards it, which is wasteful. Better:
```suggestion
        if not IsOne(topLeft^ordr) then
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+            Add(gens_bas, topLeft^ordr);
+            Add(gens_group, h^ordr);
+            # Non-constructive recognition: check whether the collected
+            # 2x2 generators already span SL(2,q).
+            if RECOG.IsThisSL2Natural(gens_bas,GF(q)) then
```
fingolfin: Isn't this test comparatively expensive? Do you really want to do this in each iteration? Besides, it is certainly pointless when `Length(gens_bas) = 1`.

Hmm... the calling code will probably want to run constructive recognition on that SL_2. So perhaps it is better to just directly call the code for that by Frank and you?  Depending on what is faster, the naming code here the constructive recognition code there (in theory of course naming should be faster, but that's only true if it is, well, true... i.e., it depends on the actual algorithms and implementation of course)

fingolfin:
```suggestion
            if RECOG.IsThisSL2Natural(gens_bas,F) then
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+    local n, F, q, one, gens_group, gens_bas, 
+          pr, t, basis1, basism1, bas, basInv,
+          h, hb, topLeft, bottomRight, ordr, K, slp, Nstart, U;
+
+    n := DimensionOfMatrixGroup(G);
+    F := FieldOfMatrixGroup(G);
+    q := Size(F);
+    one := IdentityMat(n, F);
+    Nstart := N;
+    slp := [];
```
fingolfin: Not used?
```suggestion
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+
+    if n <> 4 then
```
fingolfin: Might as well get rid of `n` and write
```suggestion
    if DimensionOfMatrixGroup(G) <> 4 then
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+        fi;
+        t := fail;
+    od;
+
+    if t = fail then
+        Print("FindSL2inSL4: out of budget while looking for a strong pre-involution\n");
+        return fail;
+    fi;
+
+    basis1 := RECOG.FixspaceMat(t);
```
fingolfin: This could be removed if you compute `basis1` already above
```suggestion
```

gap/projective/sl4_natural.gi
```diff
@@ -0,0 +1,126 @@
...
+
+    pr := ProductReplacer(GeneratorsOfGroup(G));
+
+    # find a strong pre-involution: t = x^(|x|/2) of type 2+2.
+    # RECOG.InvolutionSearcher(pr,ord,0) uses exactly one Next(pr).
+    t := fail;
+    while N > 0 do
+        t := RECOG.InvolutionSearcher(pr, Order, 0);
+        N := N - 1;
+        if t <> fail and Length(RECOG.FixspaceMat(t)) = 2 then
```
fingolfin: Does `RECOG.FixspaceMat` work right when `t` is an object-with-memory? Have you tested that case? It will matter once we are calling this from the full SL_n code.

read.g
```diff
@@ -54,6 +54,7 @@ ReadPackage("recog","gap/projective/almostsimple/lietype.gi");
 ReadPackage("recog","gap/projective/almostsimple/hints.gi");
 ReadPackage("recog","gap/projective/classicalnatural.gi");
 ReadPackage("recog","gap/projective/sl2_natural.gi");
+ReadPackage("recog","gap/projective/sl4_natural.gi");
```
fingolfin, line 57: Good, now that your functions gets loaded, have you tried actually using it, i.e., by modifying `RECOG.SLn_constructsl2` to call you code, instead?

Till-Eisen:
This function works fine. However the functions `RECOG.SLn_exceptionalgodown` and `RECOG.SLn_godownfromd` which, given a SL(d,q) as a subgroup of SL(dim, q) (in natural representation) provide a subgroup isomorphic to SL(2,q), together they cover all possible parameter q, d and dim and have a lower running time than the function `RECOG.FindSL2inSL4_natural` implemented in this PR.
Thereby either the implementation lacks efficiency or the underlying algorithm is Inferior.

PR #505, opened 2026-06-23
Fix C3 kernel seeding bug

fingolfin:
Seed trivial-image generators into the C3 kernel and add short comments explaining why both kernel-seeding sites do this.

Fixes #503

### DIFF
--- a/tst/working/quick/bugfix.tst
+++ b/tst/working/quick/bugfix.tst
@@ -423,6 +423,23 @@ true
 gap> ForAll(GeneratorsOfGroup(G), x -> SLPforElement(ri, x) <> fail);
 true
 
+# Issue #503: the C3/C5 semilinear branch must seed generators with trivial
+# field-automorphism action into the kernel, otherwise kernel recognition can
+# settle on a proper subgroup and fail final verification.
+# See https://github.com/gap-packages/recog/issues/503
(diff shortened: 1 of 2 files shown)


tst/working/quick/bugfix.tst
```diff
@@ -423,6 +423,23 @@ true
...
+gap> x2 := [ [ Z(11)^5, 0*Z(11) ], [ 0*Z(11), Z(11)^5 ] ];;
+gap> x3 := [ [ Z(11)^4, Z(11) ], [ Z(11)^6, Z(11)^9 ] ];;
+gap> G := Group([x1, x2, x3]);;
+gap> i := 35;; Reset(GlobalRandomSource, i);; Reset(GlobalMersenneTwister, i);;
+gap> ri := RecognizeGroup(G);;
+gap> IsReady(ri);
+true
+gap> Size(ri);
+24
+gap> ForAll(GeneratorsOfGroup(G), x -> SLPforElement(ri, x) <> fail);
```
Till-Eisen: Why did you additionally test `SLPforElement` here? Is the idea that a too-small 
kernel recognition would cause it to fail for some generators, making it a more sensitive check than just `Size(ri)`?

Till-Eisen (approved):

tst/working/quick/bugfix.tst
fingolfin: yeah, paranoia. Alas, we can do better now
```suggestion
gap> IsCorrect(ri);
```

PR #509, opened 2026-07-07
Speed up one test in MoreNaming.tst

fingolfin:


tst/working/combined/MoreNaming.tst
SoongNoonien, line 718: Why not here as well?

fingolfin: we could but it takes about the same either way for this small example (70 vs 60 milliseconds on my machine)

SoongNoonien (approved):

PR #516, opened 2026-07-09
Fix SL godown eigenspace dimension check

fingolfin:
This prevents non-invariant spaces from reaching LinearAction.

### DIFF
--- a/gap/projective/sl.gi
+++ b/gap/projective/sl.gi
@@ -265,7 +265,7 @@ end;
 RECOG.SLn_godownfromd:=function(g,q,d,dim)
   local y,yy,ready,order,es,dims,subsp,z,x,a,b,c,h,vec,vec2,
   pol,factors,degrees,comm1,comm2,comm3,image,basis,action,vs,readyqpl1,
-  readyqm1,count,u,orderu;
+  readyqm1,count,u,orderu,expecteddims;
 
   repeat
     ready:=false;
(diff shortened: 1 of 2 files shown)


gap/projective/sl.gi
```diff
@@ -283,7 +283,11 @@ RECOG.SLn_godownfromd:=function(g,q,d,dim)
        if not IsOne(yy) then
             es:= Eigenspaces(GF(q),yy);
             dims:=List(es,Dimension);
-            if IsSubset(Set([1,d-1,dim-d]),Set(dims)) and
+            # Since yy^(q-1)=1, yy is semisimple over GF(q).  We want
+            # one 1-space, one (d-1)-space, and the fixed outside space
+            # of dimension dim-d; the last one is absent when dim=d.
+            expecteddims:=Filtered([1,d-1,dim-d], x -> x > 0);
+            if AsSortedList(dims)=AsSortedList(expecteddims) and
                (1 in Set(dims)) then
```
TWiedemann:
```suggestion
            if AsSortedList(dims)=AsSortedList(expecteddims) then
```
1 is certainly contained in `expecteddims`, so the condition `AsSortedList(dims)=AsSortedList(expecteddims)` ensures that `1 in Set(dims)`. Hence the check `1 in Set(dims)` can be removed.

TWiedemann (approved):

tst/working/slow/bugfix.tst
```diff
@@ -26,6 +26,17 @@ gap> for i in [1..50] do
 >     ri := RECOG.TestGroup(GL(8,27), false, Size(GL(8,27)));
 > od;
 
+# The following test used to run into an error because SLn_godownfromd
+# accepted eigenspace dimensions which did not include the expected
+# fixed-space dimension.
+gap> seed:=1;;
+gap> Reset(GlobalMersenneTwister, seed);;
+gap> Reset(GlobalRandomSource, seed);;
+gap> h := GL(6,8);;
```
Till-Eisen, line 35: I could see the case $d-1 = dim-d$ as a problem in the previous code. In this case there might have been "to many" eigenspaces of dimension 1. 
Perhaps $GL(7,8)$ instead of $GL(6,8)$ would be a good test as well, since then $dim=7, d= 4$ and thereby $dim-d=d-1$?

Till-Eisen (approved):

gap/projective/sl.gi
SoongNoonien: This seems reasonable to me.

tst/working/slow/bugfix.tst
fingolfin: In fact also $dim=5$ is of interest, too, as then $dim-d=1$

PR #518, opened 2026-07-14
Move presentations for S_n/A_n to a new central file

TWiedemann:
The presentations were already there in `giant.gi`. Since they are also needed in `SnAnUnknownDegree.gi`, I moved them to a new file `presentation.gi` in `base`.

Maybe the file `presentation.gi` will contain some more presentations in the future. Maybe this will not be necessary since we have the `classicpres` package (#338), and then it might be renamed or merge into another file.

I also expanded the presentations to cover small edge cases. Currently they are not used (`giant.gi` is for `deg>=8` and `SnAnUnknownDegree.gi` is for `deg>=7`).

The tests are of a purely mathematical nature (check whether the presentations given in the paper are correct for small n, or at least yield groups with the correct size), so it is not really necessary to perform them often, but they are fast (42 ms on my laptop).


TWiedemann:
I added functions RECOG.AnNiceGens and RECOG.SnNiceGens which return a list of permutations. The presentations of S_n and A_n are given on these generators. These functions are only used in the test, but I think they might be useful.

Further, I merged the new `presentation.tst` into the old `GiantPresentation.tst`.

I expect the CI checks to pass, and then I think we can merge.

SoongNoonien (approved):

PR #519, opened 2026-07-14
Retry kernel recognition for missing generators

fingolfin:
Final verification can expose a missing residual kernel element. Add it to gensN and rerun kernel recognition before aborting.


gap/base/recognition.gi
```diff
@@ -691,33 +691,52 @@ InstallGlobalFunction( RecogniseGeneric,
...
-                " IsReady(<ri>) is not set, recognition aborts"
-            );
-            return ri;
+            # Check that all (input) generators of H lie in the group generated
+            # by the nice generators.  If a generator reveals an additional
+            # kernel element, add it and retry kernel recognition.
+            for i in [1..Length(GeneratorsOfGroup( H ))] do
+                h := GeneratorsOfGroup( H )[i];
+                if SLPforElement(ri, h) = fail then
+                    x := ImageElm(Homom(ri), h);
```
fingolfin, line 707: Hrm, I guess if this returns `fail` at this point, we are in deep troubles, and ought to abort. Actually, if we did everything right before getting here, shouldn't this be effectively impossible? Hrm.

TWiedemann: I think `x` could potentially be `fail` if `KernelRecogNode(ImageRecogNode(ri))` is too small. In any case, I agree that we ought to abort in this situation (which is what the code currently does).

TWiedemann (approved):

gap/base/recognition.gi
TWiedemann: Actually, my statement that the code currently aborts in this situation is not entirely correct. I think it does not abort in the following situation:
- The for loop finds some `h` with `SLPforElement(ri, h) = fail`, and `x := ImageElm(Homom(ri), h)` is not `fail` for this `h`, and `s := SLPforElement(ImageRecogNode(ri), x)` is also not fail. This results in `done` being set to `false`.
- In a later iteration of the for loop, we find some `h` with `SLPforElement(ri, h) = fail`, but this time `x := ImageElm(Homom(ri), h)` is `fail`. Since `done` is still `false` from a previous iteration, the code does not abort.

gap/base/recognition.gi
```diff
@@ -691,33 +691,52 @@ InstallGlobalFunction( RecogniseGeneric,
...
+                    if done then
+                        # NiceGens(ri) is too small, abort
+                        if InfoLevel(InfoRecog) = 1 and depth = 0 then
+                            Print("\n");
+                        fi;
+                        Info(InfoRecog, 1,
+                             "RecogNode <ri> could not be recognised,",
+                             " IsReady(<ri>) is not set, recognition aborts");
+                        return ri;
+                    fi;
```
TWiedemann, line 734:
```suggestion
                    abort := false;
                    if x <> fail then
                        s := SLPforElement(ImageRecogNode(ri), x);
                    else
                        s := fail;
                        abort := true;
                    fi;
                    if s <> fail then
                        y := ResultOfStraightLineProgram(s, ri!.pregensfacwithmem);
                        z := ri!.gensHmem[i] * y^-1;
                        if not isone(ri)(z) and
                           not ForAny(gensN(ri), x -> isequal(ri)(x, z)) then
                            Add(gensN(ri), z);
                            Info(InfoRecog, 2,
                                 "Final verification found extra kernel ",
                                 "element (depth=", depth, ").");
                            done := false;
                        else
                            # This should not be possible: If z is one or already in gensN(ri),
                            # then we should have been able to find an SLP for h in the first place.
                            abort := true;
                        fi;
                    fi;
                    if abort then
                        # NiceGens(ri) is too small, abort
                        if InfoLevel(InfoRecog) = 1 and depth = 0 then
                            Print("\n");
                        fi;
                        Info(InfoRecog, 1,
                             "RecogNode <ri> could not be recognised,",
                             " IsReady(<ri>) is not set, recognition aborts");
                        return ri;
                    fi;
```
I think this change makes the code abort as it should.

fingolfin:
@TWiedemann I tweaked the PR, please have a look

gap/base/recognition.gi
fingolfin: Thanks! I applied it slightly differently

TWiedemann (approved):
The change looks good to me, so I think we can merge.

